PR #41, opened 2021-04-07
ci: a minimal pipeline for the circuit

chiefbiiko:
Would basically run `cargo build` and `cargo test` for the `circuit` node when

+ pushing to `development`
+ pushing to branches that target `development` in a pull request

Defines a workflow file for just the circuit node because the [`paths`](https://github.com/t3rn/t3rn/compare/development...chiefbiiko:pipeline?expand=1#diff-051eddd49916e87d0bb5d1dd2fe0bedbada41e88871263c52801bedfd33f0bafR8) trigger filter is only available for workflows, not on job level.

### DIFF
--- a/.github/workflows/circuit_ci.yml
+++ b/.github/workflows/circuit_ci.yml
@@ -1,0 +1,44 @@
+name: Circuit CI
+
+on:
+  push:
+  pull_request:
+    branches:
+      - development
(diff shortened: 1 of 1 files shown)


MaciejBaj:
That's cool, although we opted for GitHub Actions here #49

PR #48, opened 2021-05-20
Move versatile-wasm VM to root and tailor for universal use

MaciejBaj:
- removes the dependency on pallet-contracts
- adds stack tracing
- adds additional run dedicated for Circuit
- works with FRAMEv3


MaciejBaj:
Included in #47

Issue #29, opened 2021-02-28
Receive and process execution requests from API


t3rn:
### Description

Receive execution requests from API and process them according to specified IO schedule.

### Acceptance Criteria

- [x] Create a new runtime module responsible for processing Circuit requests (a new Pallet). It will become the main Circuit engine and execution orchestrator. 
- [x] New Engine Pallet implements the `composable_exec` method for RPC API.
- [ ] Analyze received IO Schedule using regular expressions and define & document rules of passing following execution steps and phases
- [ ] New Engine Pallet is connected to Circuit node and installable as a new Pallet 
- [ ] Keep the context for the interoperable transaction
- [ ] Use off-chain workers client to dispatch the calls to external runtimes via gateways
- [ ] Analyze the response from gateways and temporarily assume it’s enough to validate the transaction’s security.

MaciejBaj:
Closed by #47

MaciejBaj:
Analyze received IO Schedule using regular expressions and define & document rules of passing following execution steps and phases was refactored by #50

PR #50, opened 2021-06-11
Refactor io-schedule parsing implementation and drop regex crate

zannis:
The regex crate does not compile in `no_std`, so this implementation drops it and uses custom parsing for io-schedule string vectors.

Also added tests for simple and complex io_schedules as well as all possible errors.


MaciejBaj (approved):
Nice 👍

MaciejBaj:
Relates to #29

Issue #27, opened 2021-01-05
Validate on-gateway execution by Circuit


MaciejBaj:
### Description

Execution performed via Gateway must be validated by Circuit. The circuit must be delivered with:
- **witness**: pre and post-execution block headers from foreign blockchain
- **execution stamp**: execution after-effects contained in execution stamp
Assuming above are collected and valid, Circuit can now re-execute the smart contract locally using [`versatile-vm`](https://github.com/t3rn/t3rn/tree/development/gateway/pallet-escrow-gateway/escrow-engine/versatile-wasm) and compare results with collected execution stamp.

### Acceptance Criteria

- [x] collect execution stamp
- [x] collect witness, block headers, from FRAME-based Substrate node and establish the transaction containing execution stamp is indeed included on that chain
- [x] implement validation 

### Additional Information

Validators authorizing execution stamps that are proven to be incorrect are not penalized **yet** for their misbehaviour. That comes with the next development milestones alongside with Proof of Stake-like system implementation.

MaciejBaj:
Closed by #55

PR #55, opened 2021-07-08
Implement Volatile VM - Relates #27

MaciejBaj:
Introduce WASM-based Volatile VM for smart contracts execution without an option of instantiation but with an option to produce execution messages that are grouped into sequential rounds.

It is pretty much the Wasm smart contracts runner introduced by `paritytech/substrate/frame/contracts`, but:
- `call`, `transfer`, `set_storage`, `get_storage` are exposed and available for external documentation
- assumes all available for execution contracts are pre-loaded into the cache. All non-preloaded contracts are assumed to be a `foreign_target` - execution won't start but call arguments are passed to `exposed call` so that the message can be created by the circuit
- can't instantiate a contract - this will be delegated to contracts repo after the whole xtx is completed
- can't charge for instantiated contract rent - tbd by the contracts repo
- extends the call stack and frames to track foreign target chain ids and enacts with round breakpoints if the chain context switches


circuit/src/execution-delivery/src/message_assembly/substrate_gateway_protocol.rs
```diff
@@ -954,11 +970,12 @@ mod tests {
             assert_signed_payload(
                 actual,
                 submitter,
-                vec![to.encode(), value.encode(), vec![]],
+                vec![to, value, vec![]],
                 vec![GatewayExpectedOutput::Events {
                     signatures: vec![b"EscrowedTransfer(address,address,value)".to_vec()],
                 }],
-                vec![1, 0, 12, 8, 4, 6, 8, 4, 1, 0],
+                // vec![1, 0, 12, 8, 4, 6, 8, 4, 1, 0],
```
MaciejBaj, line 977: @zannis different encoded call yields with different output signature

zannis: Resolved in 5fe1fb5.

zannis (approved):

circuit/src/contracts-registry/src/lib.rs
```diff
@@ -39,36 +41,47 @@ mod weights;
 pub use weights::*;
 
 pub type RegistryContractId<T> = <T as frame_system::Config>::Hash;
+pub type ChainId = [u8; 4];
```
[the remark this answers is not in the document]
MaciejBaj: Yeah, it's re-introduced in many files separately. Will extract to a separate issue alongside primitives cleanup.

PR #59, opened 2021-07-15
Refactor authority selection - Closes #60

beqaabu:
In this PR I: 

- Implemented a function called `select_authority` which changed the method of selection from selecting the first authority to selecting the authority with a given escrow account
- Added tests

PR #68, opened 2021-07-22
Refactor authority id selection - Fixes #59

beqaabu:
In this PR I added sort to the `local_keys` vector

### DIFF
--- a/circuit/src/execution-delivery/src/lib.rs
+++ b/circuit/src/execution-delivery/src/lib.rs
@@ -71,6 +71,8 @@ use crate::message_assembly::merklize::*;
 #[cfg(test)]
 mod tests;
 
+mod benchmarks;
+
 pub mod exec_composer;
 pub mod message_assembly;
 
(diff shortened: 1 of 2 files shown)


circuit/src/execution-delivery/src/lib.rs
```diff
@@ -795,7 +795,9 @@ impl<T: Config> Pallet<T> {
     }
 
     pub fn select_authority(escrow_account: T::AccountId) -> Result<AuthorityId, &'static str> {
-        let local_keys = AuthorityId::all();
+        let mut local_keys = AuthorityId::all();
+
+        local_keys.sort();
```
zannis, line 800: Fix looks fine but sorting the AuthorityId array during runtime will heavily impact weight of execution delivery as we scale upwards. At some future point we should definitely look for a way to make sure our authority ids are always sorted before invoking the `select_authority` function.

zannis (approved):
Fix looks fine, just added a comment so we keep in mind scalability of this implementation in the future.

MaciejBaj (approved):

PR #72, opened 2021-07-29
[Draft] Boilerplate code to make sure CI runst Runtime API tests

zannis:


circuit/src/contracts-registry/rpc/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use codec::Codec;
+use pallet_contracts_registry::{FetchContractsResult, RegistryContractId};
+use sp_runtime::{sp_std::vec::Vec, traits::MaybeDisplay};
+
+sp_api::decl_runtime_apis! {
+    /// The API to interact with contracts without using executive.
+    pub trait ContractsRegistryApi<AccountId, Hash> where
+        AccountId: Codec + MaybeDisplay,
+        Hash: Codec + MaybeDisplay + frame_system::pallet::Config,
```
MaciejBaj, line 11: don't think `+ frame_system::pallet::Config` is necessary here

circuit/src/contracts-registry/rpc/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+use sp_api::ProvideRuntimeApi;
+use sp_blockchain::HeaderBackend;
+use sp_core::Bytes;
+use sp_runtime::{
+    generic::BlockId,
+    traits::{Block as BlockT, MaybeDisplay},
+};
+
+#[rpc]
+pub trait ContractsRegistryApi<BlockHash, AccountId> {
```
MaciejBaj, line 21: Hash, no BlockHash

circuit/src/contracts-registry/rpc/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+            message: match e {
+                Error::DoesntExist => "Requested contract does not exist".into(),
+                Error::IsTombstone => "Requested contract is inactive".into(),
+            },
+            data: Some(format!("{:?}", e).into()),
+        }
+    }
+}
+
+impl<C, Block, AccountId> ContractsRegistryApi<Block::Hash, AccountId>
```
MaciejBaj, line 81: Hash, no BlockHash, same below. In runtime impl use Hash (frame_system::Hash) directly

MaciejBaj (changes requested):

circuit/src/contracts-registry/rpc/src/lib.rs
zannis: Seems like I can't use frame_system::Hash directly as it's private. Using frame_system::Config::Hash ends up with an 'ambiguous associated type' error.

circuit/src/contracts-registry/rpc/src/lib.rs
zannis: Will be fixed in the upcoming commit.

PR #92, opened 2021-09-02
Use live networks metadata and nonce in RPC Tests - Closes #95

### LINKED ISSUE #95: Use live networks metadata in RPC Tests
Current RPC tests use a dummy locally stored mock for metadata from XDNS to discover the correct indices for modules and methods called on remote targets. That won't work in production where RPC tests are meant to execute on mutiiple live targets (t3rn devnet, chachacha, westend, kusama etc) and each time the latest metadata of those networks must be collected. 
Create a similar test helper function to `create_test_stuffed_gateway_protocol` - `fill_gateway_protocol_with_dev_metadata`
that instead of providing mock metadata, fetched the latest metadata from the remote client. 

### Acceptance Criteria
- [ ] create new test util that fetches metadata from latest remote runtimes
- [ ] use the remote target's metadata in RPC tests

vedhavyas:
Closes #95


circuit/src/execution-delivery/Cargo.toml
```diff
@@ -88,6 +87,7 @@ hex-literal = "0.3"
 paste = "1"
 pretty_assertions = "0.7"
 wat = "1"
+relay-substrate-client = { version = "0.1.0", path = "../../../vendor/bridges/relays/client-substrate"}
```
MaciejBaj, line 90: missing `option = true`

vedhavyas: this is under dev-dependencies. Dont need have it optional there 😉

MaciejBaj (approved):

PR #163, opened 2021-11-03
Add argument for allowed side effects to register_gateway extrinsic - Closes #160

### LINKED ISSUE #160: Add argument for allowed side effects to register_gateway extrinsic
### Description

`register_gateway` function enables a sudo account to register a new gateway to circuit. Currently there's no way of specifying which extrinsic functions are callable on that target platform leaving us open to potential attack vectors. This task will introduce a new argument to the register gateway function with the list of the `SubstrateGatewayProtocol`-compatible extrinsic calls to be called by circuit during composable execution.

### Acceptance Criteria
- [ ] `register_gateway` now takes a new argument, which is a vector of strings with function names (only ones available as part of the SubstrateGatewayProtocol
- [ ] Update the XdnsRecord struct to include that new property
- [ ] `add_new_xdns_record` saves the allowed side effects field to XDNS storage
- [ ] Update register_gateway tests to include the new parameter
- [ ] Emit NewGatewayRegistered event at the end of `register_gateway`

devanshu0987:


circuit/pallets/execution-delivery/src/lib.rs
```diff
@@ -553,16 +553,17 @@ pub mod pallet {
             gateway_genesis: GatewayGenesisConfig,
             first_header: Vec<u8>,
             authorities: Option<Vec<T::AccountId>>,
+            allowed_side_effects: Option<Vec<Vec<u8>>>,
```
MaciejBaj, line 556: Introduce a type AllowedSideEffect = `Vec<u8>` and use consistently. 

Otherwise, looking good :)

MaciejBaj (changes requested):

zannis (commented):
Code changes are looking good, lets just sync up on the behaviour for onboarded chains and merge this one.

circuit/pallets/execution-delivery/src/lib.rs
zannis: Not sure about Option here, since we introduce this list, why not make it required. If not, what does 'None' mean on this field?

devanshu0987: Removed option and added new type alias.

zannis:
Some minor updates needed on benchmarks to make sure the tests pass too, please check CI.

zannis (approved):
👍🏼  ok from my side

MaciejBaj (approved):

Issue #149, opened 2021-10-26
Align licensing to t3rn source code


zannis:
### Description

Right now most of our source code includes the default Substrate GPL license which is inaccurate. We should update this to properly reflect licensing for the t3rn source code.

### Acceptance Criteria

- [ ] Agree on a license format (Apache / GPL / MIT)
- [ ] Remove existing references of Substrate licensing under t3rn-specific code
- [ ] Add license file to repo root
- [ ] Add license to the top of all source code files
- [ ] [Use rustfmt to check if source code files properly include the LICENSE file](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#license_template_path)

PR #168, opened 2021-11-10
Update licensing at main README.md

MaciejBaj:
Tiny README.md update.


zannis (commented):
Looks good but I was hoping we can use this PR to also close #149. What do you think? 

Maybe I can add a few commits on this one real quick to close them both.

PR #171, opened 2021-11-11
Generate ID for Side Effect and link with Xtx - Closes #170

### LINKED ISSUE #170: Generate ID for Side Effect and link with Xtx
### Description

Link side effects to specific Xtx. 

The id could be derived from sticking together SideEffectId and Xtx `(hash(side_effect.encode().concat(xtx.generate_id()))`, but deriving ID from a new `SideEffect` object will work just as good assuming the vectors of `FullSideEffect`will be stored in Xtx and available for lookup + update after the confirmation has been received. 

### Acceptance Criteria 
- [ ] Generate ID for AcceptedSideEffect and feed it to `confirm_side_effect` entry 
    `SideEffects::new(...args).generate_id(() { self.encode() })`
- [ ] Update Xtx with the Vector of `FullSideEffect`
- [ ] Implement `Xtx::confirm_side_effect` method that transforms the `FullSideEffect` to `CompletedSideEffect` assuming the Side Effect fits into the correct place with respec to the steps order (all side effects from the previous step must already be confirmed in order to start the new step)

MaciejBaj:
It sets out the rules for Relayers about the order of confirmation - basically, the order of the steps (schedule) needs to be respected by relayers, otherwise, Circuit won't accept the side effect confirmation (and in the future punish the relayer for not respecting the schedule)

- [x] Generate ID for side effects + unit tests
- [x] Link side effects with Xtx and implement + test their confirmation

Closes #170


circuit/primitives/src/xtx.rs
```diff
@@ -59,10 +65,458 @@ impl<
...
+                if full_side_effect.confirmed.is_none() {
+                    step_with_unconfirmed_steps = Some(i);
+                    // Recalculate the ID for each input side effect and compare with the input one.
+                    // Check the current unconfirmed step before attempt to confirm the full side effect.
+                    return if full_side_effect.input.generate_id::<Hasher>() == input_side_effect_id
+                        && step_with_unconfirmed_steps == Some(i)
+                    {
+                        // We found the side effect to confirm from inside the unconfirmed step.
+                        full_side_effect.confirmed = Some(confirmed.clone());
+                        Ok(true)
```
zannis, line 123: This result will probably depend on the contents of the ConfirmedSideEffect. The relayer might come back with some error and the SideEffect might be confirmed but unsuccessful.

circuit/primitives/src/xtx.rs
```diff
@@ -59,10 +65,458 @@ impl<
...
+            if step.is_empty() {
+                return Err("Xtx has an empty single step.");
+            }
+            for mut full_side_effect in step.iter_mut() {
+                if full_side_effect.confirmed.is_none() {
+                    step_with_unconfirmed_steps = Some(i);
+                    // Recalculate the ID for each input side effect and compare with the input one.
+                    // Check the current unconfirmed step before attempt to confirm the full side effect.
+                    return if full_side_effect.input.generate_id::<Hasher>() == input_side_effect_id
+                        && step_with_unconfirmed_steps == Some(i)
```
beqaabu, line 119: i) We know that `step_with_unconfirmed_steps` equals `Some(i)` for sure, it's assigned above, do we still need to double-check that here?
ii) Do we need variable `step_with_unconfirmed_steps` at all?

circuit/primitives/src/xtx.rs
```diff
@@ -32,12 +33,15 @@ pub struct Xtx<AccountId, BlockNumber, BalanceOf> {
 
     /// Total reward
     pub total_reward: Option<BalanceOf>,
+
+    /// Vector of Steps that each can consist out of at least one FullSideEffect
+    pub full_side_effects: Vec<Vec<FullSideEffect<AccountId, BlockNumber, BalanceOf>>>,
```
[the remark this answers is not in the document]
MaciejBaj: we can only show dependencies with a single vector assuming it comes with a mechanism in place to show dependencies (DFD). 

It's now double array to reflect that some side effects must happen in parallel vs some sequential.
- sequential: vec![vec![side_effect_0]],vec![side_effect_1]... 
- parallel: vec![vec![side_effect_0, side_effect_1...]]

circuit/primitives/src/xtx.rs
MaciejBaj: Content of ConfirmedSideEffect was checked before as a part of ConfirmExecution. 
Xtx only checks if the order of side effects is respected

circuit/primitives/src/xtx.rs
MaciejBaj: the inner loop here `Checks the current unconfirmed step before attempt to confirm the full side effect.`

var name `step_with_unconfirmed_steps` is actually confusing, would rather rename to `unconfirmed_step_no`

need to make sure that the side effects u are about to confirm is at the same `unconfirmed_step_no` and not in the future to respect the exec order

Issue #119, opened 2021-09-30
Implement post-run as the confirm side effects 


MaciejBaj:
### Description

Validate inclusion and correctness of execution for side effects and its values provided as a result of candidate for the delivered side effects results.

PR #190, opened 2021-11-26
Implement generic DFD decoder - Refers #119

MaciejBaj:
Refers #119

### DIFF
--- a/circuit/primitives/src/event_signature.rs
+++ b/circuit/primitives/src/event_signature.rs
@@ -11,57 +11,67 @@ pub fn ensure_str_err(condition: bool, err_message: &'static str) -> Result<(), 
     Ok(())
 }
 
+// Constants
+const WHITESPACE_MATRIX: [u8; 4] = [b' ', b'\t', b'\r', b'\n'];
+const ARGS_SEPARATOR: u8 = b',';
+const ARGS_START: u8 = b'(';
+const ARGS_END: u8 = b')';
+
+// Helper functions
+
+// Checks if signature is no-empty and ends correctly
+fn check_overall_sanity(signature: StrLike) -> Result<(), &'static str> {
+    let cloned = trim_whitespace(signature);
+    // make sure schedule is not empty
+    // probably irrelevant since there is already a check for that
+    let last_char = cloned.last();
+    ensure_str_err(
+        last_char.is_some(),
+        "Signature sanity failed - can't be empty",
+    )?;
+    // make sure the schedule ends correctly and remove ending character or panic
+    let ends_correctly = last_char.eq(&Some(&ARGS_END));
+    ensure_str_err(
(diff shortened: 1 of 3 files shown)


circuit/primitives/src/event_signature.rs
```diff
@@ -170,4 +237,89 @@ pub mod tests {
...
+        let valid_dfd = "(A(B(C)))";
+        // Important! If using .encode() instead of .as_bytes() + .to_vec(),
+        //  SCALE adds additional byte "92" to event name
+        let decode_res = decode_dfd(valid_dfd.as_bytes().to_vec());
+        assert_eq!(
+            decode_res,
+            Ok(vec![
+                vec![b"A".to_vec()],
+                vec![b"B".to_vec()],
+                vec![b"C".to_vec()]
```
zannis, line 305: Shouldn't this be in the reverse order?

circuit/primitives/src/event_signature.rs
```diff
@@ -170,4 +237,89 @@ pub mod tests {
...
+        let valid_dfd = "(D(C(A,B)))";
+        // Important! If using .encode() instead of .as_bytes() + .to_vec(),
+        //  SCALE adds additional byte "92" to event name
+        let decode_res = decode_dfd(valid_dfd.as_bytes().to_vec());
+        assert_eq!(
+            decode_res,
+            Ok(vec![
+                vec![b"D".to_vec()],
+                vec![b"C".to_vec()],
+                vec![b"A".to_vec(), b"B".to_vec()]
```
zannis, line 321: Same here, should be reverse

zannis (changes requested):

circuit/primitives/src/event_signature.rs
MaciejBaj: Yep!

circuit/primitives/src/event_signature.rs
MaciejBaj: For parallel events won't make a difference

circuit/primitives/src/event_signature.rs
```diff
@@ -170,4 +237,89 @@ pub mod tests {
...
+        let decode_res = decode_dfd(valid_dfd.as_bytes().to_vec());
+        assert_eq!(
+            decode_res,
+            Err("DFD Decoder - attempt to edit step at incorrect depth")
+        )
+    }
+
+    #[test]
+    fn successfully_decodes_dfd_for_3_sequential_events() {
+        let valid_dfd = "(A(B(C)))";
```
vedhavyas, line 296: can we test with some actual examples? Like transfer, swap etc... ?

MaciejBaj: added a test case for 32b long values that are IDs

Issue #180, opened 2021-11-15
Upgrade Circuit pallets to Substrate v0.9.12


MaciejBaj:
### Description

Upgrade Circuit pallets to substrate v0.9.12

### Acceptance Criteria
Upgrade pallets:
- [x] beefy-merkle-tree
- [x] beefy-mmr-pallet
- [x] contracts-registry
- [ ] execution-delivery
- [x] multi-finality-verifier
- [x] xdns

PR #191, opened 2021-11-26
Temporarily remove Exec Composer from Exec Delivery - Relates #180 

MaciejBaj:
Temporarily remove Exec Composer from Exec Delivery and associated tests + extrinsic

Relates #180


vedhavyas (dismissed):
for removing code. Anytime LGTM

PR #207, opened 2021-12-11
Add signature codec for Side Effects Raw Headers - Closes #205

### LINKED ISSUE #205: Signatures codec for side effects
### Description

Prepare the implementation for encoding and decoding signatures for side effects. This should accept a byte array as an argument and return a structure with vector of validated vector of arguments and the event name. 

The implementation would go to t3rn-primitives

Depends on #204 - extends after pallet_xnds::Config trait.

MaciejBaj:
Closes #205


zannis (commented):
Code looks good, do you mind adding some documentation on `Surrounding` and `HasXDNSAccess`?

PR #229, opened 2022-02-22
Restore pallet-evm and pallet-contracts

beqaabu:
Closes https://github.com/t3rn/x-t3rn/issues/150, https://github.com/t3rn/x-t3rn/issues/151

### DIFF
--- a/runtime/parachain/src/lib.rs
+++ b/runtime/parachain/src/lib.rs
@@ -9,7 +9,7 @@ use ethereum_light_client::EthereumDifficultyConfig;
 
 use pallet_evm::{
     EVMCurrencyAdapter, EnsureAddressNever, EnsureAddressRoot, FeeCalculator, GasWeightMapping,
-    IdentityAddressMapping, OnChargeEVMTransaction, Runner,
+    IdentityAddressMapping, OnChargeEVMTransaction, Runner, HashedAddressMapping
 };
 
 use frame_support::{
@@ -1004,16 +1004,6 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for FindAuthorTruncated<F> {
     }
 }
 
-pub struct HashedAddressMapping;
-
-impl pallet_evm::AddressMapping<AccountId> for HashedAddressMapping {
-    fn into_account_id(address: H160) -> AccountId {
-        let mut data = [0u8; 32];
-        data[0..20].copy_from_slice(&address[..]);
-        AccountId::from(Into::<[u8; 32]>::into(data))
-    }
-}
-
 impl pallet_evm::Config for Runtime {
     type FeeCalculator = FixedGasPrice;
     type GasWeightMapping = ();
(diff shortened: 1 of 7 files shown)


chiefbiiko (approved):
NICE

runtime/parachain/src/lib.rs
```diff
@@ -916,8 +927,113 @@ impl ethereum_light_client::Config for Runtime {
...
+    type Schedule = Schedule;
+    type ContractDeposit = ContractDeposit;
+    type CallStack = [pallet_contracts::Frame<Self>; 31];
+    type DeletionQueueDepth = DeletionQueueDepth;
+    type DeletionWeightLimit = DeletionWeightLimit;
+}
+
+// EVM
+parameter_types! {
+    pub const ChainId: u64 = 33;
```
zannis, line 981: can we use a const for the value here?

runtime/parachain/src/lib.rs
```diff
@@ -916,8 +927,113 @@ impl ethereum_light_client::Config for Runtime {
...
+
+pub struct HashedAddressMapping;
+
+impl pallet_evm::AddressMapping<AccountId> for HashedAddressMapping {
+    fn into_account_id(address: H160) -> AccountId {
+        let mut data = [0u8; 32];
+        data[0..20].copy_from_slice(&address[..]);
+        AccountId::from(Into::<[u8; 32]>::into(data))
+    }
+}
```
zannis, line 1015: This truncates an EVM-compatible address to return one compatible with Polkadot right? Would this mapping work?

runtime/standalone/src/lib.rs
```diff
@@ -381,26 +390,108 @@ impl t3rn_primitives::EscrowTrait for Runtime {
...
-//         pallet_evm_precompile_simple::Identity,
-//     );
-//     type ChainId = ChainId;
-//     type BlockGasLimit = BlockGasLimit;
-//     type OnChargeTransaction = ();
-//     type BlockHashMapping = pallet_ethereum::EthereumBlockHashMapping;
-// }
+// EVM
+parameter_types! {
+    pub const ChainId: u64 = 33;
```
zannis, line 395: Same as above

zannis (changes requested):
Some minor findings, otherwise nice job!

MaciejBaj (approved):

runtime/parachain/src/lib.rs
beqaabu: Yup, this is pretty much a constant which is set either directly in runtime (as [Acala](https://github.com/AcalaNetwork/Acala/blob/969878303c90f9a019df32b017e87c76c6b5d73e/runtime/acala/src/lib.rs#L1391) does for example) or set during the genesis build (like [moonbeam](https://github.com/PureStake/moonbeam/blob/7e84e15003cca1c8d9b72b8fd3e6132283f43993/node/service/src/chain_spec/moonbeam.rs#L143)). We could use either approach, but I think having it as a constant and setting it directly in the runtime config makes sense for now. Plus that number is more or less dummy until we become a parachain. 

One note would be that the `ChainId`s should be different for parachain and standalone, and I'll correct that on the following commit

runtime/standalone/src/lib.rs
beqaabu: See above

runtime/parachain/src/lib.rs
beqaabu: This mapping should work, but while looking into `pallet-evm`'s code I saw that they already have a `struct HashedAddressMapping`. Using that struct directly, instead of copying the code, would make more sense IMO. I'm gonna make the update on the following commit

MaciejBaj (approved):

PR #231, opened 2022-02-25
Clean up command.rs + fix circuit-release CI

zannis:
This PR 

- [x] cleans up some of the standalone / feature flags under node service cli
- [x] attempts to restore CI release publishing on DockerHub + GitHub

### DIFF
--- a/.github/workflows/circuit-release.yml
+++ b/.github/workflows/circuit-release.yml
@@ -5,13 +5,10 @@ on:
     branches:
       - development
     paths:
-      - ./**/*.rs
-      - ./**/Cargo.toml
-      - ./**/Cargo.lock
-      - pallets/**/*.rs
-      - pallets/**/Cargo.toml
-      - pallets/**/Cargo.lock
-      - .github/workflows/circuit-parachain-release.yml
+      - '**/*.rs'
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      - .github/workflows/circuit-release.yml
   release:
     types:
       - prereleased
@@ -53,7 +50,7 @@ jobs:
         continue-on-error: false
         with:
           command: build
-          args: --locked --manifest-path ./Cargo.toml --release
+          args: --locked --release --features with-parachain-runtime
 
       - name: 📤 Upload binary
         uses: actions/upload-artifact@v2
@@ -98,7 +95,7 @@ jobs:
           push: true
           tags: t3rn/circuit-parachain:latest
           platforms: linux/amd64
-          file: docker/circuit-parachain.Dockerfile
+          file: docker/xcm-devnet/t3rn.Dockerfile
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache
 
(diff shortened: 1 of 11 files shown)


node/cli/src/command.rs
```diff
@@ -1,168 +1,189 @@
...
+        fn impl_version() -> String {
+            "0.1.0".into()
+        }
 
-    fn description() -> String {
-        format!(
-            "Circuit Parachain Collator\n\nThe command-line arguments provided first will be \
+        fn description() -> String {
+            format!(
+                "Circuit Parachain Collator\n\nThe command-line arguments provided first will be \
```
AwesomeIbex, line 34: Is this still correct for the standalone node?

node/cli/src/command.rs
```diff
@@ -1,168 +1,189 @@
...
-fn extract_genesis_wasm(chain_spec: &Box<dyn sc_service::ChainSpec>) -> Result<Vec<u8>> {
-    let mut storage = chain_spec.build_storage()?;
-
-    storage
-        .top
-        .remove(sp_core::storage::well_known_keys::CODE)
-        .ok_or_else(|| "Could not find wasm file in genesis state!".into())
-}
-
 #[cfg(feature = "with-parachain-runtime")]
```
AwesomeIbex, line 160: I think this is definitely a good start, i didnt know we already did some of this 😄 

Although i might say this is largely one and the same still, e.g:

https://github.com/t3rn/t3rn/blob/bca51132217382092594f8beca6bf52f26588e7e/node/cli/src/command.rs#L217

and https://github.com/t3rn/t3rn/blob/bca51132217382092594f8beca6bf52f26588e7e/node/cli/src/command.rs#L25

They are the same minus the string, so im sure you can get away with something like this:
```rust
fn impl_name() -> String {
            #[cfg(feature = "with-parachain-runtime")]
            "Circuit Parachain Collator".into()
            #[cfg(feature = "with-standalone-runtime")]
            "Circuit Node".into()
}
```
Providing that we always use at least one of these features, although to be devils avocate, it does increase complexity with build tools etc.

This might help us in the future in terms of cognitive load (i think substrate loves giant files with no real module XD)

This is not for this PR, i think its just a good optimisation we can be aware of in the future as we go along, as it would be annoying to do it *all* right now

.github/workflows/circuit-release.yml
```diff
@@ -134,7 +131,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ github.event.release.upload_url }}
-          asset_path: ./target/release/parachain-collator.gz.sha256sum
+          asset_path: ./target/release/circuit-collator.gz.sha256sum
           asset_name: t3rn-circuit-parachain-collator-${{ env.PUSHED_TAG }}-x86_64-unknown-linux-gnu.gz.sha256sum
```
chiefbiiko, line 135:
```suggestion
          asset_name: circuit-collator-${{ env.PUSHED_TAG }}-x86_64-unknown-linux-gnu.gz.sha256sum
```

AwesomeIbex (approved):
LGTM!

.github/workflows/circuit-release.yml
```diff
@@ -144,6 +141,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ github.event.release.upload_url }}
-          asset_path: ./target/release/parachain-collator.gz
+          asset_path: ./target/release/circuit-collator.gz
           asset_name: t3rn-circuit-parachain-collator-${{ env.PUSHED_TAG }}-x86_64-unknown-linux-gnu.gz
```
chiefbiiko, line 145:
```suggestion
          asset_name: circuit-collator-${{ env.PUSHED_TAG }}-x86_64-unknown-linux-gnu.gz
```

.github/workflows/circuit-release.yml
zannis: Fix'd, thanks :)

node/cli/src/command.rs
zannis: Removed incorrect text, thanks

node/cli/src/command.rs
zannis: I see your point, this is how it used to be, but in this case, the SubstrateCli trait implementation would have to be the only one extracted outside the 2 modules, and then these feature flags would be needed for other functions and contstants like `load_spec`, `VERSION` etc. I feel like both versions are a bit messy and decided to keep it like this for now just to have everything grouped together.

.github/workflows/circuit-release.yml
zannis: Fixd this one too

AwesomeIbex (approved):
awesome!

MaciejBaj (approved):

PR #234, opened 2022-03-01
docs: add an initial issue template

AwesomeIbex:
Following on from https://github.com/t3rn/x-t3rn/pull/180


beqaabu (commented):
Since github supports issue forms for the public repos, would it make sense to have a separate issue template written in yaml for the t3rn repo? It would be a bit prettier and would also solve the problem with selecting the type of an issue (Epic, Story, Task, Spike) in a neat way (by creating a dropdown). Thoughts?

AwesomeIbex:
hey @beqaabu, you can see it here https://github.com/t3rn/t3rn/blob/task/make-issue-template/.github/ISSUE_TEMPLATE/backlog-item.yml

beqaabu (approved):
Amazing!

chiefbiiko (commented):
Niiice!
Do we need the top table tho, wouldn't title, label, and assignee be GitHub buttons?
And we can get rid of that .5 half point if you ask me :)

AwesomeIbex:
> Niiice! Do we need the top table tho, wouldn't title, label, and assignee be GitHub buttons? And we can get rid of that .5 half point if you ask me :)

We need them for the form :D

PR #245, opened 2022-03-17
Restore the collator release pipeline

chiefbiiko:
We were having the release pipelines attempting to run (and fail) on every ordinary push. That created a lot of clutter in our Actions. Changed the release pipeline trigger to a "pushed git tag".
Moreover, I took the freedom to reason that (in the future) we want to create binary releases for our parachain collator only (as standalone is purely for dev). Thus, the removal of the standalone release pipeline. Additionally, the collator build command assumes our `update_v0.9.17` repo structure.


.github/workflows/circuit-collator-release.yml
```diff
@@ -115,27 +103,51 @@ jobs:
...
-          > ./target/release/parachain-collator.gz.sha256sum
+          > ./target/release/circuit-collator.gz.sha256sum
 
       - name: 🏷️ Get the version tag
         run: echo "PUSHED_TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
 
+      - name: 🖨️ Get the release notes
+        run: |
+          node -e "
+            const lines = require('fs').readFileSync('./CHANGELOG.md', 'utf8').split(/(\n)/g).filter(Boolean);
```
zannis, line 122: Are we generating a CHANGELOG.md?

AwesomeIbex (dismissed):
LGTM

.github/workflows/circuit-collator-release.yml
chiefbiiko: no, just parsing it here to add it to the github release. requires to add release notes to the root changelog file before actually pushing the tag.

AwesomeIbex (approved):

MaciejBaj (approved):

PR #248, opened 2022-03-22
Add default MFV config to standalone and parachain

petscheit:
fixes circuit panic issue in MFV

Locally, I still can't build without removing the `dev-dependency` flags in pallet-circuit. Anybody facing the same?

### DIFF
--- a/node/standalone/src/chain_spec.rs
+++ b/node/standalone/src/chain_spec.rs
@@ -1,8 +1,9 @@
 use circuit_standalone_runtime::{
     AccountId, AuraConfig, BalancesConfig, ContractsRegistryConfig, GenesisConfig, GrandpaConfig,
-    MultiFinalityVerifierEthereumLikeConfig, MultiFinalityVerifierGenericLikeConfig, MultiFinalityVerifierDefaultConfig,
-    MultiFinalityVerifierPolkadotLikeConfig, MultiFinalityVerifierSubstrateLikeConfig, Signature,
-    SudoConfig, SystemConfig, XDNSConfig, WASM_BINARY,
+    MultiFinalityVerifierDefaultConfig, MultiFinalityVerifierEthereumLikeConfig,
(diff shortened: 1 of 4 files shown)


node/standalone/src/chain_spec.rs
```diff
@@ -99,7 +100,7 @@ fn fetch_xdns_record_from_rpc(
                 genesis_hash: client.genesis_hash.0.to_vec(),
             },
             gateway_sys_props,
-            vec![],
+            vec![*b"tran"],
```
chiefbiiko, line 103: is this change on purpose?

chiefbiiko (dismissed):
🎉 🎉

node/standalone/src/chain_spec.rs
petscheit: ish, in preparation for the transfer demo. But I guess we have to register the gateway manually for now, so maybe it should go

chiefbiiko: cool

runtime/parachain/src/lib.rs
```diff
@@ -509,9 +509,12 @@ construct_runtime!(
         MultiFinalityVerifierGenericLike: pallet_mfv::<Instance4>::{
             Pallet, Call, Storage, Config<T, I>
         } = 104,
-        ContractsRegistry: pallet_contracts_registry::{Pallet, Call, Config<T>, Storage, Event<T>} = 105,
-        CircuitPortal: pallet_circuit_portal::{Pallet, Call, Storage, Event<T>} = 106,
-        Circuit: pallet_circuit::{Pallet, Call, Storage, Event<T>} = 107,
+         MultiFinalityVerifierDefault: pallet_mfv::{
```
MaciejBaj, line 512: extra space here?

MaciejBaj (approved):

runtime/parachain/src/lib.rs
petscheit: cargo fmt was complaining if it left it 🤷‍♂️

PR #251, opened 2022-03-23
fix: align GatewayVendor enum layout across sub projects

chiefbiiko:
While working on the demo we were seeing `GatewayVendor=Ethereum` in Polkadot Apps UI even though in JS land we were using `GatewayVendor=Substrate`. This mini change should align GatewayVendor across the various sub projects and resolve any ambiguity.


MaciejBaj (approved):

petscheit:
Please revert! I have fixed this in the types package by switching the order there, as its less intrusive

PR #272, opened 2022-04-12
chore: rename 2 grandpa-ranger

chiefbiiko:
## Summary
Maciej came up with this epic name...


MaciejBaj (approved):

MaciejBaj:
If it was only me it would have been rang3r

PR #284, opened 2022-04-21
Rococo launch preparation

chiefbiiko:
## Summary
Bundles a wild set of changes in preparation and required for launching on Rococo.

## Checklist
- [x] adds a top level `scripts` dir
- [x]  a script that generates all required artifacts for obboarding t0rn on Rococo
- [x] defines two user space systemd services for our collators in reproducible scripts
- [x] adjusts chain spec contents, fx `RuntimeVersion.spec_name` balances, collators, sudo, etc...

Once this is merged I will need to tag development, to have chain spec changes aligned with the prebuilt collator binaries we are running and the `polkadot-js/apps` config changes I been staging.

### DIFF
--- a/node/parachain/src/chain_spec.rs
+++ b/node/parachain/src/chain_spec.rs
@@ -31,7 +31,10 @@ use log::info;
 use std::{
     convert::TryFrom,
     io::{Error, ErrorKind},
+    str::FromStr,
 };
+
+const PARACHAIN_ID: u32 = 3333_u32;
 
 /// Helper function that fetches metadata from live networks and generates an XdnsRecord
 fn fetch_xdns_record_from_rpc(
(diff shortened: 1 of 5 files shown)


node/parachain/src/chain_spec.rs
```diff
@@ -471,7 +494,74 @@ pub fn local_testnet_config() -> ChainSpec {
...
+        None,
+        // Protocol ID
+        Some("t0rn"),
+        // Fork ID
+        None,
+        // Properties
+        Some(properties),
+        // Extensions
+        Extensions {
+            relay_chain: "rococo".into(), // You MUST set this to the correct network!
```
chiefbiiko, line 563: unsure whether "rococo" is the correct literal, wasn't able to find out by looking into the code, maybe we can assert this still

MaciejBaj (dismissed):

MaciejBaj (approved):

PR #286, opened 2022-04-27
3VM: add the contract type to the contracts registry

beqaabu:
## Summary
This Adds a `ContractType` field in `ContractMetadata` and closes t3rn/3vm#50. The contract types atm could be Vanilla Evm/Wasm, Volatile Evm/Wasm, and System. 

I think storage migration for the contracts registry is not necessary for now because we don't have any real contracts in there, but it will eventually be needed in the future. I'm thinking of extracting that as a separate ticket, lmk what you think.

## Checklist


AwesomeIbex (approved):
LGTM just needs to point to 3vm_lvl2 feature branch

PR #295, opened 2022-05-05
3VM(account-manager): repatriate gas fees to contract authors

AwesomeIbex:
## Summary

These repatriates gas fees from contract authors.

It effectively adds a hook into pre_dispatch so that we can check for the call type, and if the call type is a call to the contracts pallet, we append the author data there.

Then it gets passed to the Unbalance handler in `account-manager` if the call has some beneficiary, and sends the adjusted gas fees to the beneficiary.

This should be super easy for `pallet-evm` when we do 3VM Level 3.

Due to the way that balances and `tx-payment` are designed, those traits are not in frame_support. So we need to have a macro which we define in `account-manager`, so that we don't copy-paste this. 

Also fixes some lints.

Writing tests is gonna be tricky, but let's see where we get!

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3814,7 +3814,7 @@ checksum = "437f8808009c031df3c1d532c8fd7e3d73239dfe522ebf0b94b5e34d5d01044b"
 dependencies = [
  "async-channel",
  "castaway",
- "crossbeam-utils 0.8.8",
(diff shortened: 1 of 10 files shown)


beqaabu (approved):
:weight_lifting: lgtm

PR #317, opened 2022-05-18
3vm: Provide the hardened side effects in LocalState for SDK to consume

MaciejBaj:
## Summary
As part of 3VM lvl 2 provide the hardened side effects in LocalState for SDK to consume. Side effects can be hardened since the default types between ink -> substrate are pretty much the same.

- [x] Tests have been added for bug fixes/features
- [ ] Hardened side effects are implemented as part of `t3rn-primitives`
- [ ] Updated the view in primitives::circuit::LocalCtxView to support a hardened version of primitives::volatile::LocalCtx where the side effects are concrete types.
- [x] Acceptance criteria from the issue has been completed

### DIFF
--- a/pallets/circuit/src/lib.rs
+++ b/pallets/circuit/src/lib.rs
@@ -356,11 +356,11 @@ pub mod pallet {
         }
     }
 
-    impl<T: Config> OnLocalTrigger<T> for Pallet<T> {
+    impl<T: Config> OnLocalTrigger<T, T::Escrowed> for Pallet<T> {
         fn load_local_state(
             origin: &OriginFor<T>,
             maybe_xtx_id: Option<T::Hash>,
-        ) -> Result<LocalStateExecutionView<T>, sp_runtime::DispatchError> {
+        ) -> Result<LocalStateExecutionView<T, T::Escrowed>, sp_runtime::DispatchError> {
             let requester = Self::authorize(origin.to_owned(), CircuitRole::ContractAuthor)?;
 
             let fresh_or_revoked_exec = match maybe_xtx_id {
@@ -376,9 +376,10 @@ pub mod pallet {
             )?;
             Self::apply(&mut local_xtx_ctx, None, None)?;
 
-            Ok(LocalStateExecutionView::<T>::new(
+            Ok(LocalStateExecutionView::<T, T::Escrowed>::new(
                 local_xtx_ctx.xtx_id,
                 local_xtx_ctx.local_state.clone(),
+                local_xtx_ctx.full_side_effects.clone(),
                 local_xtx_ctx.xtx.steps_cnt.clone(),
             ))
         }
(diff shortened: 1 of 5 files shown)


pallets/circuit/src/lib.rs
```diff
@@ -368,17 +369,32 @@ pub mod pallet {
...
+
+            let hardened_side_effects = local_xtx_ctx
+                .full_side_effects
+                .clone()
+                .iter()
+                .map(|step| {
+                    step.iter()
+                        .map(|fsx|
+                            HardenedSideEffect::convert(
+                                fsx.clone().harden().expect("Harden() shouldn't fail at this point after validation has been done before.")))
```
AwesomeIbex, line 389: We panic here, you can do this in multiple ways without being unsafe:
- collect results, partition and then return the errors
- use normal non-declarative flow and use error coalescence (?)

primitives/src/side_effect/mod.rs
```diff
@@ -86,22 +95,177 @@ impl Default for SecurityLvl {
...
+            ConfirmationOutcome,
+            CircuitBalance,
+            AccountId32,
+            CircuitBlockNumber,
+            CircuitBalance,
+            [u8; 4],
+            Vec<Bytes>,
+            [u8; 4],
+        ),
+        &'static str,
```
AwesomeIbex, line 125: This should really use an Error enum defined in this module, same for the other errors

primitives/src/side_effect/mod.rs
```diff
@@ -86,22 +95,177 @@ impl Default for SecurityLvl {
...
+            CircuitBlockNumber,
+            CircuitBalance,
+            [u8; 4],
+            Vec<Bytes>,
+            [u8; 4],
+        ),
+        HardenedSideEffect,
+    > for HardenedSideEffect
+{
+    fn convert(
```
AwesomeIbex, line 220: You can use the From<(..)> trait instead of custom convert

AwesomeIbex (commented):
LGTM just a few comments

pallets/circuit/src/lib.rs
MaciejBaj: yep yep will handle

primitives/src/side_effect/mod.rs
MaciejBaj: will do

primitives/src/side_effect/mod.rs
AwesomeIbex: So there are loads of implicit traits that usually imply From<..>. If we use From for all of our conversions we wont need to have any confusion. Generally whenever you need to transform some type to another you'd use From/TryFrom in rust.

MaciejBaj: done!

AwesomeIbex (approved):

PR #319, opened 2022-05-19
feat: execution signals

AwesomeIbex:
## Summary

This adds a Signal Queue and a lazy handler for applying signals.

The lazy handle could be improved in the future by sharing block space and applying weights.

This also makes side effects work for 3vm, since the hardened ones wouldn't be able to be used in SDK without a copy of the same type in SDK. We now migrate that logic to the t3rn-types crate which is strictly only types.

## Checklist

## Screenshots (if applicable)

closes https://github.com/t3rn/3vm/issues/71


pallets/circuit/src/lib.rs
```diff
@@ -356,20 +394,29 @@ pub mod pallet {
...
                 maybe_xtx_id,
             )?;
+            log::debug!(
+                target: "runtime::circuit",
+                "load_local_state with status: {:?}",
+                local_xtx_ctx.xtx.status
+            );
+
+            if maybe_xtx_id.is_none() {
+                Self::apply(&mut local_xtx_ctx, None, None)?;
```
MaciejBaj, line 418: Shouldn't load local state be clean vs on_local_trigger self::apply when no xtx found?

pallets/circuit/src/lib.rs
```diff
@@ -457,6 +515,17 @@ pub mod pallet {
 
             Ok(())
         }
+
+        fn on_signal(origin: &OriginFor<T>, signal: ExecutionSignal<T::Hash>) -> DispatchResult {
+            log::debug!(target: "runtime::circuit", "Handling on_signal {:?}", signal);
```
MaciejBaj, line 520: The rest of logs is already on ::info level, this is fine for RC

pallets/circuit/src/lib.rs
```diff
@@ -332,7 +370,7 @@ pub mod pallet {
             // Anything that needs to be done at the start of the block.
             // We don't do anything here.
             // ToDo: Do active xtx signals overview and Cancel if time elapsed
-            0
+            weight
```
MaciejBaj, line 373: create a ticket to update weight after processing the timeouts queue

types/src/abi.rs
```diff
@@ -0,0 +1,675 @@
+use codec::{Decode, Encode};
```
MaciejBaj, line 1: Is it a copy of t3rn-primitives? if so should be moved instead so there 1 implementation

MaciejBaj (dismissed):

types/src/abi.rs
AwesomeIbex: no copy, this has moved from t3rn-primitives

pallets/circuit/src/lib.rs
AwesomeIbex: we need to store the created local ctx, otherwise it doesn't exist when we come to add side effects to it, setup fails with unknownxtx

pallets/circuit/src/lib.rs
AwesomeIbex: https://github.com/t3rn/t3rn/issues/345

pallets/circuit/src/lib.rs
AwesomeIbex: Hdym? I dont think we wanna promote these to info, since they provide no value to the users

MaciejBaj (approved):

PR #355, opened 2022-06-08
Seed the XDNS registry with relevant Rococo parachains

chiefbiiko:
## Summary
Seeds the XDNS registry with all live Rococo parachains.

## Checklist
- [x] `add_new_xdns_record()` now got a `force: bool` param
- [x] `register_gateway()` now overwrites possibly existing XDNS records
- [x] adds records for all running Rococo parachains for which metadata can be obtained

## Please check that your PR completes the requirements as follows
- [x] Tests have been added for bug fixes/features
- [x] Acceptance criteria from the issue has been completed

PR #356, opened 2022-06-08
Configure grandpa-ranger and executor with Rococo parachains

chiefbiiko:
## Summary
Configures newly added XDNS subjects with the `grandpa-ranger` and `executor`. Relates to #355.

### DIFF
--- a/client/executor/config.json
+++ b/client/executor/config.json
@@ -69,5 +69,13 @@
       "rpc": "wss://rco-para.subsocial.network",
       "type": "substrate"
     }
+  ],
+  "gateways": [
+    {
+      "name": "Rococo",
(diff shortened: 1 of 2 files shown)


petscheit:
I think we should comment out the majority of configs here, as the added chains are not registered -> ranger will throw a lot of errors. Doesn't hurt to have those settings though.

chiefbiiko:
> I think we should comment out the majority of configs here, as the added chains are not registered -> ranger will throw a lot of errors. Doesn't hurt to have those settings though.

Disabled the parachain configs 4now - but kept in code as it is tedious to gather.

MaciejBaj (approved):

PR #367, opened 2022-06-16
Remove SDK from 3vm and bumps the submodule

AwesomeIbex:
## Summary
Uses sdk primitives from private registry, upgrade 3vm submodule


beqaabu (approved):
Nice! Are we also planning to move types to a separate repo and import them from the private registry?

AwesomeIbex:
> Nice! Are we also planning to move types to a separate repo and import them from the private registry?

Yessir, probably not the private registry though since it lives in public, it'll be crates.io when the documentation is written for it.

PR #374, opened 2022-06-24
Prevent replay attack when confirming side effect - Closes #366

### LINKED ISSUE #366: Prevent replay attack when confirming side effect
## Expected behavior
We should have a check in place, that ensures a side_effect was executed on target after it was created on circuit. This could be achieved by recording the last know block height of target during side_effect creation, and ensuring the side_effect was executed in a block equal or higher. This greatly reduces the attack vector (which is quite small in the first place).

### Actual behavior
Currently, when a side_effect is submitted, we do not record the last known block number of target as part of the side_effect. As a result we have no way of ensuring, a side_effect was executed after it was created. It is a small attack vector, that would allow executors to resubmit a previously executed side_effect, if the side_effect is the same. (E.g. transfer of same amount to the same person in two different side_effects)

MaciejBaj:
## Summary
- [ ] Expose read and compare best available height from Portal
- [ ] Store submission_target_height in encoded form to FullSideEffect at validation 
- [ ] Compare submission with confirmation height at confirmation
- [ ] Brute seed block height 1 to circuit unit tests mock GRANDPA Finality Verifier 

- [x] Tests have been added for bug fixes/features
- [x] Acceptance criteria from the issue has been completed

### DIFF
--- a/pallets/circuit-portal/src/xbridges/mod.rs
+++ b/pallets/circuit-portal/src/xbridges/mod.rs
@@ -3,6 +3,7 @@ use codec::{Decode, Encode};
 
 use sp_core::crypto::ByteArray;
 use sp_finality_grandpa::SetId;
+use sp_runtime::traits::Header;
 
 use sp_std::vec::Vec;
 use sp_trie::{read_trie_value, LayoutV1, StorageProof};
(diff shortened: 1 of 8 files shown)


pallets/circuit-portal/src/xbridges/mod.rs
```diff
@@ -62,9 +65,45 @@ pub fn get_roots_from_bridge<T: pallet_multi_finality_verifier::Config<I>, I: 's
             gateway_block_hash,
         )
         .ok_or(Error::<T>::StepConfirmationBlockUnrecognised)?;
+
     Ok((extrinsics_root, storage_root))
 }
 
+pub fn read_cmp_latest_height_from_bridge<
```
petscheit, line 72: Looks good. How would we go about exposing this via RPC? The ranger will also need access to this, to enable the `catch_up` logic we spoke about.

petscheit (dismissed):
Looks good!

pallets/circuit-portal/src/xbridges/mod.rs
MaciejBaj: Added to `read latest gateway height` RPC Runtime and RPC. Also cleaned up unused execute composable step via Portal

AwesomeIbex (approved):
lgtm

petscheit (approved):

PR #376, opened 2022-06-29
Remove Ranger Rust Decoder

petscheit:
## Summary
The ranger was so far relying on a rust script that was used to decode the justification. We now know how to decode in TS, so the rust script is obsolete. This PR removes it from the repo


chiefbiiko:
Maybe also rm at client/xtx/justification-decoder

MaciejBaj (dismissed):

petscheit:
I've removed the decoder also from `client/xtx`, as pointed out by @chiefbiiko

chiefbiiko (approved):
LGTM

MaciejBaj (approved):

PR #391, opened 2022-07-12
fix: Correct received event names after new SFX are available for executors

petscheit:
The SideEffect naming has changed, this updates the executor accordingly

### DIFF
--- a/client/executor/circuit/listener.ts
+++ b/client/executor/circuit/listener.ts
@@ -1,4 +1,3 @@
-// import "@t3rn/types"
 import { EventEmitter } from "events"
 import { ApiPromise, WsProvider } from "@polkadot/api"
 import { SideEffect } from "../utils/types"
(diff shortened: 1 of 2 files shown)


client/executor/index.ts
```diff
@@ -1,3 +1,4 @@
+require('dotenv').config()
```
chiefbiiko, line 1: on purpose?

chiefbiiko (dismissed):
LGTM

client/executor/index.ts
petscheit: yup, because I work with a local .env file. Shouldn't interfere with anything else

client/executor/circuit/listener.ts
```diff
@@ -1,4 +1,4 @@
-import "@t3rn/types"
+// import "@t3rn/types"
```
MaciejBaj, line 1: remove

MaciejBaj (changes requested):

MaciejBaj (approved):

PR #394, opened 2022-07-13
Confirm source for escrowed security to all standard SFX

MaciejBaj:
## Summary
Confirm source for escrowed security to all standard SFX

## Checklist


pallets/xdns/src/lib.rs
```diff
@@ -427,6 +436,15 @@ pub mod pallet {
             )
         }
 
+        fn get_gateway_security_coordinates(chain_id: &ChainId) -> Result<Bytes, &'static str> {
+            if !<XDNSRegistry<T>>::contains_key(chain_id) {
+                return Err("Xdns record not found while accessing security coordinates")
+            }
+            Ok(<XDNSRegistry<T>>::get(chain_id)
+                .expect("Ensured the XDNSRegistry contains given chain id")
```
AwesomeIbex, line 444: I would say we should try to reduce panicking the worker when we can emit an error message/return the error.

In general, panic bad, result good. Especially if we are already returning the result so can be as easy as `ok_or_else(|_| Err("Ensured the XDNSRegistry contains given chain id"))?`


Although another argument can be made that we should return an `Error` enum for this pallet, but that's an optimisation for now.

petscheit (commented):
some small things

petscheit (dismissed):

pallets/xdns/src/lib.rs
MaciejBaj: addressed

petscheit (approved):

AwesomeIbex (approved):

PR #400, opened 2022-07-14
Utilise the rearchitecture for 3VM and install EVM to the nodes

AwesomeIbex:
## Summary

This adds the supplementary changes to support 3vm lvl 3. As well as importing the new functionality into the runtime for parachain and standalone. At the moment EVM is configured in parachain but disabled via origin.

This allows account manager to return a generated ID for use later.
Fixes a bug where it would assume there is always a beneficiary.
Bumps SDK primitives to the minor version.
Uses a contract type for the type in ContractMetadata (why are these all blobs? can we be smarter with this? The size is unbounded).
Add's 3VM traits.
Adds a very noddy configuration of EVM, needs work (ticket in backlog)


runtime/standalone/src/contracts_config.rs
```diff
@@ -59,21 +79,146 @@ impl pallet_3vm_contracts::Config for Runtime {
...
     type Randomness = RandomnessCollectiveFlip;
     type Schedule = Schedule;
-    type SignalBounceThreshold = ConstU32<5>;
+    type ThreeVm = ThreeVm;
     type Time = Timestamp;
     type WeightInfo = pallet_3vm_contracts::weights::SubstrateWeight<Self>;
     type WeightPrice = pallet_transaction_payment::Pallet<Self>;
 }
+
+// impl pallet_evm::Config for Runtime {
```
MaciejBaj, line 97: Pls add a todo on when planning to resolve it

primitives/src/account_manager.rs
```diff
@@ -42,7 +42,11 @@ pub enum Reason {
 
 pub trait AccountManager<Account, Balance> {
     /// Send funds to a recipient via the escrow
-    fn deposit(payee: &Account, recipient: &Account, amount: Balance) -> DispatchResult;
+    fn deposit(
+        payee: &Account,
+        recipient: &Account,
+        amount: Balance,
+    ) -> Result<u64, sp_runtime::DispatchError>;
```
MaciejBaj, line 49: What's the returned value here?  extend the docs

MaciejBaj (changes requested):

AwesomeIbex:
ci is failing because of XBI but my branch doesn't have the pallet :/ seems like a race condition on CI runner

MaciejBaj (approved):

PR #415, opened 2022-08-01
Prepare continous collator deployments

chiefbiiko:
## Summary
Reintroduces a collator release pipeline step that publishes a collator Docker image to the GitHub container registry. Additionally, the pipeline now triggers a continuous deployment pipeline in our chainops repo.

### DIFF
--- a/.github/workflows/circuit-collator-release.yml
+++ b/.github/workflows/circuit-collator-release.yml
@@ -7,15 +7,15 @@ on:
 
 env:
   PARACHAIN_NAME: t0rn
-  RUST_BACKTRACE: 1
+  RUST_BACKTRACE: full
   CARGO_TERM_COLOR: always
 
 jobs:
   build-release:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: ☁ Checkout git repo
-        uses: actions/checkout@master
+        uses: actions/checkout@v3
         with:
           submodules: recursive
           token: ${{ secrets.GH_PAT }}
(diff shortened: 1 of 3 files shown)


collator.Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM phusion/baseimage:jammy-1.0.0
```
AwesomeIbex, line 1: Would it not be easier to use alpine or something more lightweight? Do we need ubuntu?

This would mean we can use official images too.

.github/workflows/circuit-collator-release.yml
```diff
@@ -53,8 +53,43 @@ jobs:
...
+      - name: 📥 Download circuit collator
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ github.sha }}-collator-release-artifacts
+          path: ./target/release/
+
+      - name: 🏷️ Get the version tag
+        run: echo "PUSHED_TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+
+      - name: 🐋 Login to Docker Hub
```
AwesomeIbex, line 74: As far as im aware, dockerhub rate limits pulls, so if we try to scale, we might fail to scale. Another issue is a would-be attacker could just pull our docker image as a bot and effectively denial of service. The alternative here is ghcr(https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)

collator.Dockerfile
chiefbiiko: alpine and musl would make things more complicated

.github/workflows/circuit-collator-release.yml
chiefbiiko: your scenario isn't valid here - this pipeline only runs when we push a git tag

AwesomeIbex: That's not exactly true. If our git flow is: push image tag > trigger deployment. Then a third party can block that deployment easily as above.

AwesomeIbex (approved):

MaciejBaj (approved):

PR #416, opened 2022-08-03
feat: Calculate claims for settled Xtx with Executors, AccountManager and GlobalClock 

MaciejBaj:
## Summary
Calculate claims for settled Xtx for Executors and Stakers with pallets `Executors`, `AccountManager` and newly introduced `GlobalClock`  - refer to README for more context.

## Checklist
- [ ] feat: Account Manager handles xtx rewards refunds and implelemnts `on_claimable`
- [ ] feat: Scaffold of global circuit clock collecting claimables
- [ ] feat: Extend Executors and AccountManager interfaces with claims
- [ ] chore: Handle extra collators + staker roles in Treasury
- [ ] feat: Plug accounts manager to Circuit] 
- [ ] refactor: Rename pallet-staking to pallet-executors
 

pallets/account-manager/src/manager.rs
```diff
@@ -100,6 +196,74 @@ impl<T: Config> AccountManagerExt<T::AccountId, BalanceOf<T>> for Pallet<T> {
 
         Ok(())
     }
+
+    /// Collect claimable (only SFX execution rewards) for Executors and Stakers submitted by Circuit at the duration of the current Round
+    fn on_collect_claimable(
```
AwesomeIbex, line 201: i think this should just take all ledger items and claim them when the storage items are better generalised. really this pallet shouldn't care about who is who, I think.

Or probably this should have a todo to move the executor specific logic to pallet-executors and remove the storage items. but really that will be more hassle with storage migrations than to just add a pallet and remove those coupled items.

pallets/circuit-clock/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        type Currency: ReservableCurrency<Self::AccountId>;
+
+        type WeightInfo: weights::WeightInfo;
+
+        /// Type providing some time handler
+        type Time: frame_support::traits::Time;
+
+        #[pallet::constant]
+        type EscrowAccount: Get<Self::AccountId>;
```
AwesomeIbex, line 67: hmm, should the clock have access to the escrow account? I don't think this is used here.

pallets/circuit-clock/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+        circuit_clock::{CircuitClock, ClaimableArtifacts},
+        transfers::EscrowedBalanceOf,
+    };
+
+    #[pallet::config]
+    pub trait Config: frame_system::Config {
+        /// The overarching event type.
+        type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+        type Currency: ReservableCurrency<Self::AccountId>;
```
AwesomeIbex, line 59: Take currency from ::Escrowed

pallets/account-manager/src/manager.rs
```diff
@@ -46,6 +59,89 @@ impl<T: Config> AccountManagerExt<T::AccountId, BalanceOf<T>> for Pallet<T> {
         Ok(())
     }
 
+    /// Reward executor for successful sfx execution - accounted after successful xtx resolution
+    fn commit_charge(executor: T::AccountId, sfx_id: T::Hash) -> DispatchResult {
+        let pending_charge: SfxRequestCharge<T::AccountId, BalanceOf<T>> =
+            if let Some(pending_charge) =
+                SfxPendingChargesPerRound::<T>::get(T::CircuitClock::current_round(), sfx_id)
```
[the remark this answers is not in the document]
MaciejBaj: `SfxPendingChargesPerRound` converts to `SfxSettlement` only after externally called by Circuit. This must only happen after the Xtx is successfully resolved - if COMMIT then call for `commit_charge`, if REVERT then `refund_charge`. 

The funds transfer to executor is designed to happen asynchronously only at the time executor / staked is ready to `claim` - global clock must implement it (todo). That's a good business value proposition since is in the full control of when would the transfers appear in the context of taxable events etc.

pallets/circuit-clock/src/lib.rs
MaciejBaj: yeah we'd need to decide what accounts distribute the rewards at claim. for now T::EscrowAccount keeps the rewards until the claim, not the biggest fan

pallets/account-manager/src/manager.rs
```diff
@@ -46,6 +59,89 @@ impl<T: Config> AccountManagerExt<T::AccountId, BalanceOf<T>> for Pallet<T> {
...
+        T::Currency::reserve(&charge.requester, charge.charge_fee + charge.offered_reward)?;
+
+        SfxPendingChargesPerRound::<T>::insert(
+            T::CircuitClock::current_round(),
+            sfx_id,
+            charge.clone(),
+        );
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
MaciejBaj: asynchronous deposits explained above

pallets/account-manager/src/manager.rs
```diff
@@ -46,6 +59,89 @@ impl<T: Config> AccountManagerExt<T::AccountId, BalanceOf<T>> for Pallet<T> {
         Ok(())
     }
 
+    /// Reward executor for successful sfx execution - accounted after successful xtx resolution
+    fn commit_charge(executor: T::AccountId, sfx_id: T::Hash) -> DispatchResult {
```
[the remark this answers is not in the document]
MaciejBaj: this is meant to essentially only be called by Circuit at the event of Xtx resolution with either COMMIT or REVERT (after msbhv or timeout). do u see it implemented in Circuit / somewhere else then account manager? the charges and rewards from the contracts should also be aligned with the async model btw

pallets/account-manager/src/manager.rs
MaciejBaj: agreed, will expose another method `fn return_claims()` from AccountManager

pallets/account-manager/src/manager.rs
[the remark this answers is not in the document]
AwesomeIbex: So do we not just add another Reason to the finalize function? since that is essentially all that it does

pallets/account-manager/src/manager.rs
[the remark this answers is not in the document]
AwesomeIbex: Ah so for point one, perhaps that should be some storage in the circuit, then and only once it is ready to be committed to the account manager, you should move it as a ledger item to be paid.

This does a few things:
- avoids inconsistency with data
- decouples the account manager from circuit and keeps the responsibility of only managing stuff that is going to be paid
- avoids polluting the account manager with very specific storage items

Another option, since i see your point about having some "promotable" ledger items, that is, an item that should not be issues until some rule.

In which case, this should try to make this less about circuit and more about the domain of this pallet, maybe PendingLedgerItem. Which would be functionality that allows a transfer to be promoted to the queue and thus handled by the clock.

pallets/account-manager/src/lib.rs
```diff
@@ -88,6 +108,44 @@ pub mod pallet {
...
+    >;
+
+    #[pallet::storage]
+    pub type TotalSfxRewardCountPerRound<T: Config> = StorageMap<
+        _,
+        Blake2_128,
+        RoundInfo<T::BlockNumber>,
+        <T::Currency as Currency<T::AccountId>>::Balance,
+    >;
+
```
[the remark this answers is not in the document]
MaciejBaj: `totals` aren't used anywhere and should be removed. 

We should for sure just go with one type, I wasn't diving too deep into refactoring the contracts charge logic - this is now open for a discussion, but should IMO 
1. needs to follow the same flow of pre-charge (reserve) -> refund / commit as other SFXs - after the Xtx resolution with either COMMIT or REVERT signal 
2. follow asynchronous payments to contract builders via `circuitClock::claim` and produce the `ClaimableArtifacts` with `BenefitSource::ContractRewards / BenefitSource::ContractInflation (for pallet treasury) / BenefitSource::ContractStaking`
3. pick one of `ExecutionRegistryItem` / `SfxSettlement` and refactor the account manager accordingly.

primitives/src/circuit_clock.rs
```diff
@@ -0,0 +1,37 @@
+use codec::{Decode, Encode};
+use frame_support::dispatch::DispatchResult;
+use scale_info::TypeInfo;
+use sp_runtime::RuntimeDebug;
+use sp_std::{fmt::Debug, prelude::*};
+use crate::common::RoundInfo;
+
+#[derive(Clone, Eq, PartialEq, PartialOrd, Encode, Decode, RuntimeDebug, TypeInfo)]
+pub enum CircuitRole {
+    Ambassador,
```
chiefbiiko, line 10: would get rid of `Ambassador` for the sake of simplicity at the beginning

primitives/src/circuit_clock.rs
```diff
@@ -0,0 +1,37 @@
...
+    Executor,
+    Staker,
+    Collator,
+    ContractAuthor, // Builders
+    Relayer,
+    Requester,
+    Local,
```
chiefbiiko, line 17: who are `Requester` and `Local`?

pallets/treasury/src/lib.rs
```diff
@@ -371,7 +383,7 @@ pub mod pallet {
 
             // allocate to beneficiary
             T::Currency::deposit_into_existing(&who, BalanceOf::<T>::from(total_rewards))
-                .expect("Should deposit balance to account"); //TODO
+                .expect("Should deposit balance to account"); // TODO
```
chiefbiiko, line 387: i believe
+ we should introduce the factory https://docs.rs/pallet-treasury/3.0.0/pallet_treasury/
+ refactor the inflation mechanism in the existing pallet-treasury stub back to sth like pallet-inflation
+ while keeping things in line with the new circuit timing pallet

chiefbiiko (commented):
suggest we merge this branch to development - in order to gradually build on top

MaciejBaj:
Closing the draft. Will follow up with a series of PRs introducing separately:
- clock 
- claim 
- account manager adapted for asynchronous claims
- staking pallet -> executors pallet

PR #430, opened 2022-08-24
[DRAFT] Align side effects to SDK

AwesomeIbex:
## Summary

This enables the full e2e submission of SDK side effects as of `0.1.0`. At the moment we can only submit 1 due to some bug in SDK [here](https://github.com/t3rn/sdk/issues/17)

## Checklist
- [ ] moves `harden` to use novel conversion traits
- [ ] formats on circuit
- [ ] fix a bug where if side effect hardening would fail then the apply would've happened beforehand, now we make sure all fallible work first
- [ ] adds balance to LocalState that is shared with SDK, now we don't need hardened versions of LocalCtx
- [ ] be verbose about where the circuit logs so we don't need to have runtime spam when debugging
- [ ] clean up the nested inner functions to use simple boolean variables instead of functions

### DIFF
--- a/pallets/circuit/Cargo.toml
+++ b/pallets/circuit/Cargo.toml
@@ -31,8 +31,8 @@ sp-trie    = { default-features = false, git = "https://github.com/paritytech/su
 
 t3rn-primitives     = { path = "../../primitives", default-features = false }
 t3rn-protocol       = { path = "../../protocol", default-features = false }
-# t3rn-sdk-primitives = { version = "0.1.0", default-features = false }
-t3rn-sdk-primitives = {  path = "../../../sdk/crates/primitives", default-features = false }
+t3rn-sdk-primitives = { version = "=0.1.1-rc.4", default-features = false }
+#t3rn-sdk-primitives = {  path = "../../../sdk/crates/primitives", default-features = false }
 
 orml-traits = { git = "https://github.com/t3rn/open-runtime-module-library", branch = "polkadot-v0.9.19", default-features = false }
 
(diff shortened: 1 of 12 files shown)


pallets/circuit/Cargo.toml
```diff
@@ -31,7 +31,8 @@ sp-trie    = { default-features = false, git = "https://github.com/paritytech/su
 
 t3rn-primitives     = { path = "../../primitives", default-features = false }
 t3rn-protocol       = { path = "../../protocol", default-features = false }
-t3rn-sdk-primitives = { version = "0.1.0", default-features = false }
+# t3rn-sdk-primitives = { version = "0.1.0", default-features = false }
+t3rn-sdk-primitives = {  path = "../../../sdk/crates/primitives", default-features = false }
```
AwesomeIbex, line 35: todo release

pallets/circuit/src/lib.rs
```diff
@@ -1752,30 +1773,32 @@ impl<T: Config> Pallet<T> {
...
+                        SecurityLvl::Dirty
                     }
-                    SecurityLvl::Dirty
                 }
-                let submission_target_height = T::CircuitPortal::read_cmp_latest_target_height(
-                    side_effect.target,
-                    None,
-                    None,
-                )?;
+                // let submission_target_height = T::CircuitPortal::read_cmp_latest_target_height(
```
AwesomeIbex, line 1787: todo reenable

AwesomeIbex:
Merging since i assume the protocol issue is something fixed by further improvements and shouldn't block the merge into this feature branch

PR #476, opened 2022-10-13
Connect pallet Assets as Foreign Fungibles Transactor to XCM

MaciejBaj:
## Summary
Connect pallet Assets as Foreign Fungibles Transactor to XCM

## Checklist

### DIFF
--- a/runtime/t0rn-parachain/src/xcm_config.rs
+++ b/runtime/t0rn-parachain/src/xcm_config.rs
@@ -64,9 +64,9 @@ pub type ForeignFungiblesTransactor = FungiblesAdapter<
     Assets,
     // Use this currency when it is a fungible asset matching the given location or name:
     ConvertedConcreteAssetId<
-        AssetId,
+        parachains_common::AssetId,
         Balance,
-        AsPrefixedGeneralIndex<AssetsPalletLocation, AssetId, JustTry>,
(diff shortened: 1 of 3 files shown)


runtime/t0rn-parachain/src/xcm_config.rs
```diff
@@ -52,6 +56,32 @@ pub type LocalAssetTransactor = CurrencyAdapter<
...
 
+// The non-reserve fungible transactor type
+// It will use pallet-assets, and the Id will be matched against AsAssetType
+// This is intended to match FOREIGN ASSETS
+pub type ForeignFungiblesTransactor = FungiblesAdapter<
+    // Use this fungibles implementation:
+    Assets,
+    // Use this currency when it is a fungible asset matching the given location or name:
+    ConvertedConcreteAssetId<
+        AssetId,
```
AwesomeIbex, line 67: Is this assetId correct? it can't be an enum I think. I was having issues getting it to work with enums so had to set it as u32

AwesomeIbex (commented):
LGTM one question

AwesomeIbex (approved):

runtime/t0rn-parachain/src/xcm_config.rs
MaciejBaj: yeah replaced with `common_parachains::AssetId` for transacting which also is u32 https://github.com/paritytech/cumulus/blob/master/parachains/common/src/lib.rs#L65

PR #485, opened 2022-10-23
Intro shell zombienet

chiefbiiko:
## Summary

**tl;dr** `./zombienet.sh`

Introduces a `rococo-local` `zombienet` running a shell runtime equivalent.

For configuration of another chain regenerate the spec using the collator's `build-spec` command, possibly edit it, and save it as `zombienet.json`.

## Checklist
- [x] Runs a `rococo-local` with two t3rn shell parachains
- [x] HRMP channels setup


MaciejBaj (dismissed):

zombienet.toml
```diff
@@ -0,0 +1,53 @@
...
+command = "circuit-collator"
+ws_port = 9931
+
+[[hrmp_channels]]
+sender = 3000
+recipient = 3001
+max_capacity = 8
+max_message_size = 512
+
+[[hrmp_channels]]
```
AwesomeIbex, line 49: needs bidirectional channels here

zombienet.toml
```diff
@@ -0,0 +1,53 @@
...
+[[relaychain.nodes]]
+name = "alice"
+ws_port = 9900
+
+[[relaychain.node_groups]]
+name = "bob"
+count = 4
+
+[[parachains]]
+id = 3000
```
AwesomeIbex, line 14: we should map them to the digit in the port number so its less confusing, so we know based on the port number than 9920 is 2000

zombienet.sh
```diff
@@ -0,0 +1,34 @@
...
+root_dir=$(git rev-parse --show-toplevel)
+
+mkdir -p /tmp/zombienet/bin
+
+if [[ ! -x /usr/local/bin/zombienet ]]; then
+  curl -fL# -o /usr/local/bin/zombienet https://github.com/paritytech/zombienet/releases/download/$version/zombienet-$machine
+  chmod +x /usr/local/bin/zombienet
+fi
+
+if [[ ! -x /tmp/zombienet/bin/polkadot ]]; then
```
AwesomeIbex, line 24: we should just pull the binaries into .bin and ignore them, so we don't have to pull polkadot/polkadot-parachain every time

zombienet.sh
```diff
@@ -0,0 +1,34 @@
...
+  tmp_dir=$(mktemp -d)
+  git clone --branch $pdot_branch --depth 1 https://github.com/paritytech/polkadot $tmp_dir
+  cargo build --manifest-path $tmp_dir/Cargo.toml --features fast-runtime --release --locked
+  mv -f $tmp_dir/target/release/polkadot /tmp/zombienet/bin/polkadot
+fi
+
+cargo build --manifest-path $root_dir/node/$runtime/Cargo.toml --release --locked
+cp -f $root_dir/target/release/circuit-collator /tmp/zombienet/bin/circuit-collator
```
AwesomeIbex, line 32: this wouldnt work no? zombienet expects the binary in the path, I would suggest making this a diff command and letting users setup their path vs building every time just to spawn.

AwesomeIbex (changes requested):

AwesomeIbex:
> For configuration of another chain regenerate the spec using the collator's `build-spec` command and/or edit `zombienet.json`.

What does this mean? Zombienet generates a spec every time itself unless you use a custom command in the file.

chiefbiiko:
> > For configuration of another chain regenerate the spec using the collator's `build-spec` command and/or edit `zombienet.json`.
> 
> ~What does this mean? Zombienet generates a spec every time itself unless you use a custom command in the file.~
> 
> Why do we do custom chainspecs?

Our t3rn runtime doesn't have `rococo-local` preconfigured, that's why had to intro the custom spec.

zombienet.sh
chiefbiiko: Rite now it will do a full build on first spawn only, then rustc got our back.

AwesomeIbex:
> > > For configuration of another chain regenerate the spec using the collator's `build-spec` command and/or edit `zombienet.json`.
> > 
> > 
> > ~What does this mean? Zombienet generates a spec every time itself unless you use a custom command in the file.~
> > Why do we do custom chainspecs?
> 
> Our t3rn runtime doesn't have `rococo-local` preconfigured, that's why had to intro the custom spec.

i used `rococo` for all of mine and it works, no need for custom chainspecs

zombienet.sh
AwesomeIbex: > Rite now it will do a full build on first spawn only, then rustc got our back.

It doesn't, it fights with the ide if it changes the lockfile so rebuilds every time. Also polkadot-parachain-runtime builds every time because its a sys and non incremental

MaciejBaj (changes requested):

MaciejBaj (dismissed):
⭐

PR #491, opened 2022-11-02
Update sfx_id hashing logic to prevent collisions

petscheit:
## Summary
We still had an issue with side-effects potentially having hash collisions when created. This is now fixed, by using H(xtx_id, sfx_index) to generate the ID. XTX_ID is always unique, as its generated with H(address, nonce). This way we can't run into hash collisions anymore.

A SFX hash can now be generated with the FullSideEffect or SideEffect struct. Typically we either pass the array of side effects, or a single full side effect around, so this setup enables the id to be generated in both cases (FSX stores the sfx index)

## Checklist
- [x] move nonce fields from SideEffect and to FullSideEffect as `index`
- [x] implement hashing as described above
- [x] refactor confirm_sfx extrinsic, removing unneeded parameters, reducing tx costs
- [x] rename bid_execution to bid_sfx and remove needed parameter, reducing tx cost
- [x] fix all tests


pallets/circuit/src/lib.rs
```diff
@@ -744,7 +745,7 @@ pub mod pallet {
             max_exec_cost: u128,
             max_notifications_cost: u128,
         ) -> DispatchResultWithPostInfo {
-            let sfx_id = side_effect.generate_id::<SystemHashing<T>>();
+            let sfx_id = side_effect.generate_id::<SystemHashing<T>>(xtx_id.as_ref(), 0u32);
```
petscheit, line 748: Im setting sfx_index to 0 here because only one SFX is passed. Is that correct?

MaciejBaj (approved):

AwesomeIbex (dismissed):

pallets/circuit/src/lib.rs
AwesomeIbex: For now yes, this is YAGNI and not used so will need changing probably later.

pallets/circuit/src/lib.rs
```diff
@@ -1249,10 +1249,6 @@ impl<T: Config> Pallet<T> {
                     .flat_map(|(cnt, fsx_step)| {
                         fsx_step
                             .iter()
-                            .map(|full_side_effect| full_side_effect.clone())
```
petscheit: @MaciejBaj I removed the local filter here, because I use the mapping for lookups in confirm and bid

pallets/circuit/src/tests.rs
```diff
@@ -3897,6 +3959,7 @@ fn execute_side_effects_with_xbi_works_for_call_evm() {
                 xtx_id,
                 valid_evm_sfx
                     .generate_id::<circuit_runtime_pallets::pallet_circuit::SystemHashing<Runtime>>(
```
AwesomeIbex, line 3961: Could we have a test to cover the collision please?

Something like 2 Xtx's, different id, same indices on side effect, hash their ids, prove different.

petscheit: Done! (fingers crossed the pipeline passes on Friday evening right before I pack up)

AwesomeIbex (approved):

MaciejBaj (approved):

Issue #403, opened 2022-07-15
Enforce encodedArg parameter length as defined in GatewayABIConfig


petscheit:
## Expected behavior
When creating a SideEffect, it should be ensured that the encodedArgs have the length described in the gatewayAbi. An error should be thrown if the length is incorrect.

### Actual behavior
currently, we don't check if the encodedArgs length matches the one specified in the gateway configuration. This can lead to unwanted behavior. For example, codec converts as ss58 address into a 46 bytes (I think) bytes array by default. This then leads to the address being cut of, making it impossible to release the funds at the end of the lifecycle. Essentially, they are now bricked.

This can be ensured by checking that the encoded length matches the one that is stored in `GatewayAbiConfig`. The obvious spot to enforce this would be `types/src/abi.rs:eval_abi()`. If the length check fails, a fitting error should be returned in that function. This automatically reverts the transaction.

AwesomeIbex:
I'd be curious to know the breadth of this encoding issue. If it is true that storing ss58s is problematic, this issue extends beyond circuit and everywhere we use T::AccountId in a storage item.

PR #502, opened 2022-11-09
Enforce encoded parFix for issue #403

palozano:
## Summary
- When creating Sideeffect, enforce encoded arguments' length as defined in `GatewayABIConfig` by checking that the (encoded) length matches the one in GatewayAbiConfig. 
- If the length check fails, a fitting error should be returned in that function.
- This automatically reverts the transaction.

## Checklist
- [ ] Changed code from `eval_abi()` function within `types/src/abi.rs`. 

- [ ] Tests have not been added for bug fixes/features.
- [ ] Acceptance criteria from the issue has been completed.


types/src/abi.rs
```diff
@@ -258,28 +258,41 @@ impl Type {
         gen: &GatewayABIConfig,
```
AwesomeIbex, line 258: May we have tests to support these in case of any regressions?

With bugs, it's essential to prove that the bug is fixed via a test.

types/src/abi.rs
```diff
@@ -258,28 +258,41 @@ impl Type {
...
-                },
-                32 => {
-                    let res: [u8; 32] = decode_buf2val(encoded_val)?;
-                    Ok(res.encode())
-                },
-                _ => unimplemented!(),
+            Type::Address(size) => {
+                if size != &gen.address_length {
+                    Err("Encoded address has different length than the gateway's configuration")
+                } else {
```
AwesomeIbex, line 264: If i were to optimise this, i'd add a function on the enum to try and check this before I do anything, this check would simply either pass if the length is the same as configured or be ignored if it doesn't apply (e.g: dynamic address)

This avoids any Christmas tree like stuff where we repeat the same error

AwesomeIbex (commented):
Looks good, just need some tests and a nitpick

types/src/abi.rs
palozano: Added test as part of new commit.

types/src/abi.rs
palozano: Refactored as a method on the enum as part of the new commit.

AwesomeIbex (dismissed):

PR #506, opened 2022-11-14
Removes Insurance event and connects emit to on_init hook

### LINKED ISSUE #504: Xtx kill functionality doesn't clean up state properly
When a XTX is dropped at bidding, the state must be cleaned up, and the user refunded. This is done in `on_initialize`, which checks for XTXs that need to be dropped/killed. The detection of these works, however the state is not cleaned up properly.

The state is cleaned up with `fn kill()` in circuit. The bug arises in that function. 
```

fn kill(local_ctx: &mut LocalXtxCtx<T>, cause: CircuitStatus) {
        local_ctx.xtx.status = cause.clone();
        Optimistic::<T>::try_slash(local_ctx);

        Self::square_up(local_ctx, None)
            .expect("Expect Revert and RevertKill options to square up to be infallible");

        Self::apply(local_ctx, (cause.clone(), cause));
    }

```
We pass the same cause to apply, which in turn doesn't match correctly. Typically, we generate the `status_change` in the update function, which should generate source and target of every state transition. When using kill, this doesn't happen, however. As a result, XTXs that should be removed from storage are killed every block, never being removed from storage.

In general, this is caused by inconsistencies in the implementation. Usually `fn update()` is used for generating the status update `(old_status, new_status)`, which is then passed to `fn apply()` to perform the actual state updates. 

We should consider a couple of things to ensure this situation can't arise again in the future:
- ensure `fn apply()` has no wildcard in the matching
- use `fn update()` to generate the status updates for every state transition
- revamp the circuit testing suite

on branch `fix/kill-state-cleanup` I have left a failing test that outlines the issue

petscheit:
- removes obsolete `SideEffectInsuranceReceived` event, as it's no longer needed.
- separates `fn emit()` and connects event emission to `on_initialize`
- temporary fix #504


pallets/circuit/src/lib.rs
```diff
@@ -375,6 +375,7 @@ pub mod pallet {
                         &Self::account_id(),
                         Zero::zero(),
                         Some(xtx_id),
+                        // ToDo consider moving xtx_bids to xtx_ctx in order to self update to always determine status
```
MaciejBaj, line 378: Move the conditional check against pending xtx bid timeout

MaciejBaj (changes requested):

pallets/circuit/src/lib.rs
MaciejBaj: other than that good to go

MaciejBaj (approved):

PR #509, opened 2022-11-17
Install a standard FRAME treasury

chiefbiiko:
## Summary
Installs a FRAME treasury to all runtimes to facilitate transparent on-chain spending.

## Details

The `Treasury::account_id()` is based upon pot seed `PalletId(*b"pottrsry")`

Because of the configured `SpendOrigin`

``` rs
impl pallet_treasury::Config for Runtime {
  /// The origin required for approving spends from the treasury outside of the proposal
  /// process...
  type SpendOrigin = NeverEnsureOrigin<Balance>;
```

actual treasury spending requires setup of the FRAME governance pallets that enable on-chain proposals...

## Checklist

- [x] All runtimes have a treasury
- [x] `pallet-clock` now implements `current_round() -> RoundInfo<T::BlockNumber>`
- [x] Removes `pallet-executors` from the workspace as it is outdated and depends on our old treasury's `current_round()`


pallets/clock/src/lib.rs
```diff
@@ -15,12 +15,11 @@ pub use t3rn_primitives::{
     abi::{GatewayABIConfig, Type},
     account_manager::AccountManager,
     claimable::ClaimableArtifacts,
-    clock::Clock,
+    clock::Clock as TClock,
```
AwesomeIbex, line 18: Not really sure why we do this here, we should aim for consistency and only override the trait if needs be. In this case I don't see anything called clock imported so probably a little pedantic to change this. There's probably a style rule somewhere on this too.

pallets/executors/src/lib.rs
```diff
@@ -38,14 +38,15 @@ pub mod pallet {
     };
     use sp_std::collections::btree_map::BTreeMap;
     use t3rn_primitives::{
+        // treasury::Treasury as TTreasury,
+        clock::Clock as TClock,
```
AwesomeIbex, line 42: Same here

runtime/t0rn-parachain/src/lib.rs
```diff
@@ -152,7 +152,8 @@ construct_runtime!(
         // Monetary stuff.
         Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>} = 10,
         TransactionPayment: pallet_transaction_payment::{Pallet, Storage, Event<T>} = 11,
-        Assets: pallet_assets::{Pallet, Call, Storage, Config<T>, Event<T>} = 12,
```
AwesomeIbex, line 155: can we keep this 12 please? these indices shouldnt be ad volatile as this.

AwesomeIbex (dismissed):

AwesomeIbex (approved):

PR #518, opened 2022-11-25
Fix/bidding errors

petscheit:
## Summary
Fixes two issues that became apparent during bidding

1) adds getBondValue function, which either returns the latest bid or, if no bid is found, the max_reward. This removes the deadlock that a multi-step XTX would be initiated in. Before, bidding was not possible as a panic arose

2) adds try_dropped_at_bidding_refund, which takes care of refunding the executors that have bid on the SFXs of the dropped XTX. Before, they would be slashed which is incorrect.


pallets/circuit/src/lib.rs
```diff
@@ -1441,7 +1441,11 @@ impl<T: Config> Pallet<T> {
 
     fn kill(local_ctx: &mut LocalXtxCtx<T>, cause: CircuitStatus) {
         local_ctx.xtx.status = cause.clone();
-        Optimistic::<T>::try_slash(local_ctx);
+        if let CircuitStatus::RevertTimedOut = cause {
```
MaciejBaj, line 1444: prefer match cause for CircuitStatus::DroppedAtBidding =>  Optimistic::<T>::try_dropped_at_bidding_refund(local_ctx); and Optimistic::<T>::try_slash(local_ctx) for everything else

MaciejBaj (changes requested):

pallets/circuit/src/lib.rs
petscheit: very true

petscheit:
There was another issue with an XTX not being removed from the timeout map. This resulted in the circuit trying to revert a confirmed transaction, which causes some issues, in this case, a panic that stopped block production. I've applied a fix that resolves the issue. Nothing to fancy, as we're rebuilding these parts anyways

MaciejBaj (approved):

Issue #510, opened 2022-11-18
Multi-currency reward payments for Executors 


MaciejBaj:
## Description
Extend max_reward and bid to work with more than a single currency - base currency. 

Integrate Circuit, Account Manager and Treasury to work one of the base currencies - subset of assets registered in pallet-assets. 

Implement adapter enacting common balance ops with both NativeCurrency and Assets:

```rust
pub struct Monetary<AccountId, Assets, NativeCurrency>(
    PhantomData<(AccountId, Assets, NativeCurrency)>,
);
impl<AccountId, Assets: Balanced<AccountId>, NativeCurrency: ReservableCurrency<AccountId>>
    Monetary<AccountId, Assets, NativeCurrency>
{
    pub fn deposit(
        beneficiary: &AccountId,
        asset_id: Option<Assets::AssetId>,
        amount: NativeCurrency::Balance,
    ) where
        Assets::Balance: From<NativeCurrency::Balance>,
    {
        match asset_id {
            None => {
                NativeCurrency::deposit_creating(beneficiary, amount);
            },
            Some(asset_id) => {
                Assets::deposit(asset_id, beneficiary, amount.into()).unwrap();
            },
        }
    }
}
```
### Acceptance Criteria

- [ ] executors handle SFX in ROC on Rococo

PR #526, opened 2022-11-27
Enable SFX rewards, bidding and slashing in multiple assets

MaciejBaj:
Enable SFX rewards, bidding and slashing in multiple assets

Solves: https://github.com/t3rn/t3rn/issues/510

## Checklist

- [x] extend SFX interface with reward asset id field
- [x] add optional asset id to AccountManager::deposit
- [x] create new submodule to AccountManager::Monetary that performs the same withdraw and deposit operations on both NativeCurrency and Assets 
- [x] extend AccountManager interface with can_withdraw, withdraw_immediately , deposit_immediately 
- [x] manager submodule at AccountManager doesn't reserve anymore - use direct withdraw and deposit  
- [x] use Monetary module to withdraw and deposit back requester funds at SFX order request
- [x] switch AccountManager to use Monetary
- [x] implement executors bidding for SFX in foreign assets using Monetary 
- [x] implement executors slashing in foreign assets using Monetary 
- [x] only allow execution with single foreign assets for all SFX


pallets/account-manager/src/lib.rs
```diff
@@ -69,6 +78,10 @@ pub mod pallet {
 
         #[pallet::constant]
         type EscrowAccount: Get<Self::AccountId>;
+
+        type AssetBalanceOf: Convert<BalanceOf<Self>, AssetsBalanceOf<Self>>;
```
AwesomeIbex, line 82: nice

pallets/account-manager/src/manager.rs
```diff
@@ -205,12 +226,14 @@ impl<T: Config> AccountManagerExt<T::AccountId, BalanceOf<T>, T::Hash, T::BlockN
...
-                maybe_actual_fees
-            ).expect("Expect try finalize to be infallible");
+            <Self as AccountManagerExt<
+                T::AccountId,
+                BalanceOf<T>,
+                T::Hash,
+                T::BlockNumber,
+                <T::Assets as Inspect<T::AccountId>>::AssetId,
+            >>::finalize(charge_id, outcome, maybe_recipient, maybe_actual_fees)
+            .expect("Expect try finalize to be infallible");
```
AwesomeIbex, line 236: `try` being infallible is a bit of misnomer 🙊

pallets/account-manager/src/manager.rs
```diff
@@ -275,6 +298,38 @@ impl<T: Config> AccountManagerExt<T::AccountId, BalanceOf<T>, T::Hash, T::BlockN
...
+        amount: BalanceOf<T>,
+        asset_id: Option<<T::Assets as Inspect<T::AccountId>>::AssetId>,
+    ) -> DispatchResult {
+        Monetary::<T::AccountId, T::Assets, T::Currency, T::AssetBalanceOf>::withdraw(
+            payee, amount, asset_id,
+        )
+    }
 }
 
 #[cfg(test)]
```
AwesomeIbex, line 335: may we have a test to cover some of these applied to an asset?

pallets/account-manager/src/monetary.rs
```diff
@@ -0,0 +1,203 @@
...
+        beneficiary: &AccountId,
+        asset_id: Option<Assets::AssetId>,
+        amount: NativeCurrency::Balance,
+    ) {
+        match asset_id {
+            None => {
+                NativeCurrency::deposit_creating(beneficiary, amount);
+            },
+            Some(asset_id) => {
+                Assets::deposit(asset_id, beneficiary, AssetBalanceOf::convert(amount)).unwrap();
```
AwesomeIbex, line 32: We shouldn't panic here, we should also make sure we prove the asset exists in or before the call to deposit and return an error if it doesn't or we cant.

AwesomeIbex (changes requested):
Looks great! Just some small comments

pallets/account-manager/src/manager.rs
MaciejBaj: yea I suggest we fix this part as Circuit rewrite to state machine

pallets/account-manager/src/monetary.rs
MaciejBaj: done

pallets/circuit/src/lib.rs
```diff
@@ -1655,6 +1633,14 @@ impl<T: Config> Pallet<T> {
...
             })?;
 
+            if index < side_effects.len() - 1
+                && side_effects[index].reward_asset_id != side_effects[index + 1].reward_asset_id
+            {
+                return Err(
+                    "SFX validate failed - enforce all SFX to have the same reward asset field",
+                )
+            }
+
```
AwesomeIbex, line 1643: can we not do a safe index check by
- fold over each side effect, initial being true
- if asset id != next asset id, set to false
- variable is assets_are_same

This avoids looping each check and can be an easy testable terminal op

MaciejBaj: changed to using .get 
            if let Some(next) = side_effects.get(index + 1) {


not in favor of extra loop since we already enumerate over SFX array

petscheit (dismissed):
Sick stuff, I like.

AwesomeIbex (dismissed):

petscheit (dismissed):

petscheit (approved):

PR #528, opened 2022-11-29
Enforce encoded arguments' length

### LINKED ISSUE #403: Enforce encodedArg parameter length as defined in GatewayABIConfig

palozano:
- refactor: update pallets account-manager src
- refactor: update circuit for safe math
- refactor: update executors for safe math
- refactor: update treasury for safe math
- refactor: update xdns for safe math
- test: update account-manager for safe math test
- chore: missed an update on circuit for safe math test
- chore: missed an update on treasury for safe math test
- fix: update safe math err handling
- chore: update safe math for err handling
- chore: update safe math for err handling in circuit
- chore: update safe math

When creating SideEffect, enforce encoded arguments' length as defined 
in GatewayABIConfig by checking that the (encoded) length matches the 
one in GatewayAbiConfig.

If the length check fails, a fitting error is returned in that function,
reverting the transaction.

Fixes #403

### DIFF
--- a/pallets/circuit/src/optimistic/mod.rs
+++ b/pallets/circuit/src/optimistic/mod.rs
@@ -40,8 +40,11 @@ impl<T: Config> Optimistic<T> {
             }
         }
         // Check if bid candidate has enough balance and reserve
-        <T::Escrowed as EscrowTrait<T>>::Currency::reserve(executor, bid + fsx.input.insurance)
-            .map_err(|_e| Error::<T>::BiddingRejectedExecutorNotEnoughBalance)?;
+        <T::Escrowed as EscrowTrait<T>>::Currency::reserve(
+            executor,
+            bid.checked_add(&fsx.input.insurance).unwrap(),
+        )
+        .map_err(|_e| Error::<T>::BiddingRejectedExecutorNotEnoughBalance)?;
 
         let mut sfx_bid =
             SFXBid::<T::AccountId, EscrowedBalanceOf<T, T::Escrowed>>::new_none_optimistic(
@@ -58,9 +61,12 @@ impl<T: Config> Optimistic<T> {
         // Un-reserve the funds of discarded bidder.
         // Warning: From this point on all of the next operations must be infallible.
         if let Some(current_best_bid) = &current_accepted_bid {
-            let mut total_unreserve = current_best_bid.insurance + current_best_bid.bid;
+            let mut total_unreserve = current_best_bid
+                .insurance
+                .checked_add(&current_best_bid.bid)
+                .unwrap();
             if let Some(bond) = current_best_bid.reserved_bond {
-                total_unreserve += bond;
+                total_unreserve = total_unreserve.checked_add(&bond).unwrap();
             }
             <T::Escrowed as EscrowTrait<T>>::Currency::unreserve(
                 &current_best_bid.executor,
@@ -118,7 +124,7 @@ impl<T: Config> Optimistic<T> {
 
                 <<T as Config>::Escrowed as EscrowTrait<T>>::Currency::unreserve(
                     &sfx_bid.executor,
-                    insurance + reserved_bond,
+                    insurance.checked_add(&reserved_bond).unwrap(),
                 );
             }
         }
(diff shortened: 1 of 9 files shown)


AwesomeIbex:
this is safemathops no?

pallets/circuit/src/optimistic/mod.rs
```diff
@@ -40,7 +40,12 @@ impl<T: Config> Optimistic<T> {
             }
         }
         // Check if bid candidate has enough balance and reserve
-        <T::Escrowed as EscrowTrait<T>>::Currency::reserve(executor, bid + fsx.input.insurance)
+        let _bid = if let Some(v) = bid.checked_add(&fsx.input.insurance) {
```
MaciejBaj, line 43: no underscore since used

pallets/circuit/src/optimistic/mod.rs
```diff
@@ -116,17 +132,22 @@ impl<T: Config> Optimistic<T> {
                 let (insurance, reserved_bond) =
                     (*sfx_bid.get_insurance(), *sfx_bid.expect_reserved_bond());
 
+                let nsrnc = if let Some(v) = insurance.checked_add(&reserved_bond) {
```
MaciejBaj, line 135: rename to let checked_insurance =

pallets/circuit/src/optimistic/mod.rs
```diff
@@ -147,10 +168,24 @@ impl<T: Config> Optimistic<T> {
                 };
 
                 // First slash executor
-                slashed_reserve += insurance + reserved_bond;
+                if let Some(v) = slashed_reserve
+                    .checked_add(&insurance)
+                    .unwrap()
```
MaciejBaj, line 173: no unwrap

pallets/treasury/src/lib.rs
```diff
@@ -304,23 +305,54 @@ pub mod pallet {
...
             let (count_devs, count_execs) =
-                <Beneficiaries<T>>::iter_keys().fold((0, 0), |mut acc, c| {
+                <Beneficiaries<T>>::iter_keys().fold((0, 0), |mut acc: (i64, i64), c| {
                     match c.1 {
                         BeneficiaryRole::Developer => {
-                            acc.0 += 1;
+                            if let Some(v) = acc.0.checked_add(1) {
+                                acc.0 = v
+                            }
+                            // cannot return Err
```
MaciejBaj, line 314: create todos and open corresponding issue

pallets/treasury/src/lib.rs
```diff
@@ -304,23 +305,54 @@ pub mod pallet {
...
-            let relative_per_exec =
-                Perbill::from_rational(1, count_execs as u32) * inflation_alloc.executor;
+            let relative_per_dev = if let Some(v) =
+                Perbill::from_rational(1, count_devs as u32).checked_mul(&inflation_alloc.developer)
+            {
+                v
+            } else {
+                Perbill::from_rational(1, count_devs as u32)
+            };
+            // cannot return Err
```
MaciejBaj, line 340: create todos and open corresponding issue

MaciejBaj (changes requested):

pallets/circuit/src/optimistic/mod.rs
palozano: Done

pallets/circuit/src/optimistic/mod.rs
palozano: replaced it with two if let statements

pallets/treasury/src/lib.rs
palozano: Added the TODOs and the issue

pallets/treasury/src/lib.rs
palozano: Added the TODOs and the issue

PR #529, opened 2022-11-30
Safe math operations

palozano:
- refactor: update pallets account-manager src
- refactor: update circuit for safe math
- refactor: update executors for safe math
- refactor: update treasury for safe math
- refactor: update xdns for safe math
- test: update account-manager for safe math test
- chore: missed an update on circuit for safe math test
- chore: missed an update on treasury for safe math test
- fix: update safe math err handling
- chore: update safe math for err handling
- chore: update safe math for err handling in circuit
- chore: update safe math
- fix: restore cargo lock file
- fix: restore again cargo lock

## Summary

Refactored raw math operations across t3rn/pallets/*,
and 3vm/pallets/3vm.

##Details

Raw math replaced by checked_* operations to return errors or
default values.

Division by zero has been secured.

Tests with math incorporate checked_* ops now.


pallets/circuit/src/lib.rs
```diff
@@ -439,7 +442,13 @@ pub mod pallet {
                             &vec![],
                             None,
                         );
-                        deletion_counter += 1;
+                        if let Some(v) = deletion_counter.checked_add(1) {
+                            deletion_counter = v;
+                        }
+                        // cannot return the error
```
AwesomeIbex, line 448: comment here

AwesomeIbex: Just make sure we log and short circuit via `return`

pallets/circuit/src/lib.rs
```diff
@@ -1297,10 +1317,15 @@ impl<T: Config> Pallet<T> {
                     local_ctx.xtx_id,
                     local_ctx.xtx.timeouts_at,
                 );
-                <PendingXtxBidsTimeoutsMap<T>>::insert::<XExecSignalId<T>, T::BlockNumber>(
-                    local_ctx.xtx_id,
-                    T::SFXBiddingPeriod::get() + frame_system::Pallet::<T>::block_number(),
-                );
+                // cannot return an error if checked_add fails
+                if let Some(v) = T::SFXBiddingPeriod::get()
```
AwesomeIbex, line 1321: same here

pallets/circuit/src/lib.rs
```diff
@@ -1877,9 +1909,15 @@ impl<T: Config> Pallet<T> {
 
                     queue.swap_remove(0);
 
-                    remaining_key_budget -= 1;
-                    // apply has 2
-                    processed_weight += db_weight.reads_writes(2 as Weight, 1 as Weight);
+                    if let Some(v) = remaining_key_budget.checked_sub(1) {
+                        remaining_key_budget = v
+                    } // cannot return an error, signature is Weight
```
AwesomeIbex, line 1914: same for all of these

pallets/treasury/src/lib.rs
```diff
@@ -304,23 +305,54 @@ pub mod pallet {
...
             let (count_devs, count_execs) =
-                <Beneficiaries<T>>::iter_keys().fold((0, 0), |mut acc, c| {
+                <Beneficiaries<T>>::iter_keys().fold((0, 0), |mut acc: (i64, i64), c| {
                     match c.1 {
                         BeneficiaryRole::Developer => {
-                            acc.0 += 1;
+                            if let Some(v) = acc.0.checked_add(1) {
+                                acc.0 = v
+                            }
+                            // cannot return Err
```
AwesomeIbex, line 314: and again, just check all the hooks and do the above comment

AwesomeIbex (changes requested):
Good stuff, just some dilligence around logging and a way to short circuit the issue if it cannot be returned

pallets/circuit/src/lib.rs
AwesomeIbex: also ideally try do the check higher up so we don't make any storage changes and fail validation at the end

palozano: done

pallets/circuit/src/lib.rs
palozano: How can I short-circuit in this case where I have a fn with return type Weight? Should I change it or just log::error!() and skip the loop?

pallets/circuit/src/lib.rs
palozano: (Same as previous)

pallets/circuit/src/lib.rs
AwesomeIbex: just log and return the current weight counter, or you could break the loop ya

AwesomeIbex (approved):

PR #534, opened 2022-12-03
Refactor Circuit as a state machine

MaciejBaj:
### Acceptance Criteria 
- [x] Implement Circuit as a state machine as outlined 
- [ ] Transition from SideEffects array to Xtx is tested separately for the setup stage 
- [x] Kill is tested and infallible for all combinations CircuitStatus 
- [x] Compile is tested and produces expected state transitions for all combinations of CircuitStatus
- [x] Circuit as state machine is integrated and used by all current pallet-circuit extrinsics: 
  - [x] on_extrinsic_trigger 
  - [x] submit_bid
  - [x] on_initialized-> kill tick 
  - [x] cancel_xtx 
  - [x] on_initialized-> collect_bids tick
  - [x] on_initialized-> process_signal (3vm) tick
  - [x] confirm_side_effect
  - [x] execute_side_effects_with_xbi -> do_xbi_exit


pallets/circuit/src/lib.rs
```diff
@@ -372,53 +374,38 @@ pub mod pallet {
...
+                            // Account fees and charges
+                            Self::square_up(local_ctx, status_change, None).expect(
+                                "Expect square up at DroppedAtBidding loop to be infallible",
+                            );
                             Self::emit_status_update(
                                 local_ctx.xtx_id,
-                                Some(local_ctx.xtx),
+                                Some(local_ctx.xtx.clone()),
                                 None,
                             );
```
AwesomeIbex, line 405: I think we need to move the `can_withdraw` step into `compile`. This way, then and only then, when all validation steps pass, can we apply the function to commit.

So compile or something before should ensure that ANYTHING that can fail can pass, then we apply F.

To do this, I suggest:
- remove result return time from square up
- `PSFX::get()` -> ensure it exists first in compile and just unwrap/expect
- check status is not equal to pendingbidding first

You can provide some functor which would do this validation first, if you don't want to hide it in `compile`.

So two HOF's:
- one to validate
- then to commit the changes after compile

pallets/circuit/src/lib.rs
```diff
@@ -372,53 +374,39 @@ pub mod pallet {
...
+                                    // Error - at least one FSX has no bid
+                                    return PrecompileResult::TryKill(Cause::Timeout)
+                                }
+                            }
+                            PrecompileResult::UpdateFSX(current_fsx.clone())
+                        },
+                        |status_change, local_ctx| {
+                            // Account fees and charges
+                            Self::square_up(local_ctx, status_change, None).expect(
+                                "Expect square up at DroppedAtBidding loop to be infallible",
```
MaciejBaj, line 400: todo: separate square up to fallible locks and infallible unlocks

pallets/circuit/src/lib.rs
MaciejBaj: we can split the current square up operations to fallible and infallible:

- A) fallible lock requester for all SFX max rewards @CircuitStatus::Requested
- B) fallible lock executor at bidding for SFX @CircuitStatus::PendingBidding | @CircuitStatus::InBidding
- C) fallible executor's attempt to execute SFX via XBI @CircuitStatus::PendingBidding

- D)  (expected to be infallible) unreserve requester's max rewards and reserve requester's rewards for bid amounts @CircuitStatus::Ready
- E) infallible unreserve requester's max rewards and executor's bid amounts @CircuitStatus::Killed(_)
- F) infallible unreserve requester's max rewards and slash dishonest executors @CircuitStatus::Revert(_)
- G) infallible rewards payouts via AccountManager::finalize and infallible unlock executor's bonds @CircuitStatus::Finalize

I could move infallible D-G as part of compile / apply as they're all infallible unreserve / finalize 

then separate fallible A-C into another module and use separate by fn on_extrinsics_trigger, fn execute_sfx_via_xbi, fn bid_sfx in pre-compilation validation calls

pallets/circuit/src/lib.rs
```diff
@@ -372,53 +374,39 @@ pub mod pallet {
...
+                            }
+                            for mut fsx in current_fsx.iter_mut() {
+                                let sfx_id = fsx.generate_id::<SystemHashing<T>, T>(xtx_id);
+                                // Either assign best bid to FSX or Kill entire Xtx with DroppedAtBidding cause.
+                                if let Some(best_sfx_bid) = <PendingSFXBids<T>>::get(xtx_id, sfx_id)
+                                {
+                                    fsx.best_bid = Some(best_sfx_bid);
+                                } else {
+                                    // Error - at least one FSX has no bid
+                                    return PrecompileResult::TryKill(Cause::Timeout)
```
petscheit, line 392: Doesnt this lead to slashing again, even though we just refund on droppedAtBidding?

AwesomeIbex:
I don't think we should take the delta approach in this case as its just too confusing, just lazily apply anything that can fail upwards and inject what to do when something doesn't fail. This should be fairly straightforward vs. collecting all the deltas for state transitions and committing them.

petscheit:
I agree that doing this delta approach with storage/local_ctx updates is a lot. Feel like it would be quite a hot setup, be a very valid question is ifs worth the work and the potential complexity it might bring. So yeah, strike that for storage updates. Also very valid that they might fail.

I do think we should consider dealing with non-local_ctx updates differently, though. If we look at things like account manager, events, slashing, etc. It doesn't seem complicated to implement this delta approach, and I think two big benefits are testing and extensibility. I kind of see that we'll fall into the same trap again when we add new functionality and extend the state machine with new tx types etc. Correct me if I'm wrong, but I don't see a massive simplification of how testable this is compared to the current state.

AwesomeIbex:
Lazy evaluation is probably the most testable. State deltas require mocks which are written by people piped into some function.

This simply passes a list of things into `compile` which may fail, then and only then does it store state.

A simple test might be:
`compile(|| Error(()), |res| { if res.is_ok() { update_state() }})`

Did state update? no because I passed an error

Did state update if I pass no errors? damn sure

Wb a black box test? Pass all the things, did state update?

This is a much easier control flow than producing a large object of state deltas which are mocked into compile

pallets/circuit/src/lib.rs
```diff
@@ -375,53 +378,39 @@ pub mod pallet {
...
```
petscheit, line 396: Isn't this the wrong cause?

pallets/circuit/src/lib.rs
```diff
@@ -375,53 +378,39 @@ pub mod pallet {
...
-
-                    let current_step = local_ctx.xtx.steps_cnt.0;
-                    for mut fsx in local_ctx.full_side_effects[current_step as usize].iter_mut() {
-                        let sfx_id = fsx.generate_id::<SystemHashing<T>, T>(local_ctx.xtx_id);
-                        if let Some(best_sfx_bid) = <PendingSFXBids<T>>::get(xtx_id, sfx_id) {
-                            fsx.best_bid = Some(best_sfx_bid);
-                        } else {
-                            // error - some FSX don't have bids
-                            Self::kill(&mut local_ctx, CircuitStatus::DroppedAtBidding);
+                    Machine::<T>::compile_infallible(
```
petscheit, line 381: Can we move this to the machine maybe so it's coherent with the revert logic below?

pallets/circuit/src/lib.rs
```diff
@@ -646,61 +623,52 @@ pub mod pallet {
...
+                },
+                no_post_updates,
+            )?;
 
-            Self::kill(&mut local_ctx, CircuitStatus::DroppedAtBidding);
+            Ok(().into())
+        }
 
+        #[pallet::weight(<T as pallet::Config>::WeightInfo::cancel_xtx())]
+        pub fn revert(origin: OriginFor<T>, xtx_id: T::Hash) -> DispatchResultWithPostInfo {
```
petscheit, line 647: Wouldnt this brick funds?

pallets/circuit/src/lib.rs
```diff
@@ -1430,51 +1103,30 @@ impl<T: Config> Pallet<T> {
...
-            _ => {},
-        }
-
-        Self::square_up(local_ctx, None)
-            .expect("Expect Revert and RevertKill options to square up to be infallible");
-
-        Self::apply(local_ctx, (cause.clone(), cause));
-    }
-
     fn square_up(
```
petscheit, line 1120: This function im a bit dubious of. Wouldn't it make sense to tie the different operations/updates directly into the transitions of the state machine? The we would only keep the transitions there in mind when updating and we can modularize the logic here.

pallets/circuit/src/machine/mod.rs
```diff
@@ -0,0 +1,516 @@
...
+                <pallet::Pallet<T> as Store>::XExecSignals::mutate(local_ctx.xtx_id, |x| {
+                    *x = Some(local_ctx.xtx.clone())
+                });
+                // Always clean temporary PendingSFXBids and TimeoutsMap after bidding
+                <pallet::Pallet<T> as Store>::PendingSFXBids::remove_prefix(local_ctx.xtx_id, None);
+                <pallet::Pallet<T> as Store>::PendingXtxBidsTimeoutsMap::remove(local_ctx.xtx_id);
+
+                true
+            },
+            (_, CircuitStatus::Killed(_cause)) => {
```
petscheit, line 443: How do you feel about defining these without wildcards? So this one should be `PendingBidding | InBidding` or something like that?

pallets/circuit/src/machine/mod.rs
```diff
@@ -0,0 +1,516 @@
...
+                        );
+                    }
+                }
+                // Always clean temporary PendingSFXBids and TimeoutsMap after bidding
+                <pallet::Pallet<T> as Store>::PendingSFXBids::remove_prefix(local_ctx.xtx_id, None);
+                <pallet::Pallet<T> as Store>::PendingXtxBidsTimeoutsMap::remove(local_ctx.xtx_id);
+
+                true
+            },
+            (_, CircuitStatus::Reverted(_cause)) => {
```
petscheit, line 462: Same here

pallets/circuit/src/machine/test.rs
```diff
@@ -0,0 +1,238 @@
+#[cfg(test)]
+pub mod test {
```
petscheit, line 2: I like this stuff; looking so much cleaner already, wow. Nice with these guard tests, what I'm missing a bit is that we test every transition for correct but also incorrect transitions. Just to be sure that we cover everything, and no faulty transitions are possible. Maybe it makes sense to have a single test for each transition (e.g. kill) and then run the test from each available state, the allowed an forbidden ones.

The other thing would be to track an XTX through the different paths, and having them non-empty. But maybe that more for circuit tests?

petscheit (commented):
Sick stuff, I like where this going a lot. Left some smaller comments, mainly about testing and how to progress there.

MaciejBaj:
continued in #562

PR #552, opened 2022-12-13
Add git hooks

palozano:
To address some of the problems we've been having lately, I realized git hooks exist and we could use two of them to format the commit messages and the code (so CI is happy when pushing commits).

This PR adds two git hooks:
	1. Check if the commit you write follows the conventional
	   commits rules.
	2. Run `cargo fmt` and `cargo fix --allow-dirty --allow-staged`
	   to format accordingly to rustc.

It's necessary to run `git config core.hooksPath .githooks` in the repo to get the (new) hooks.
If one wants to bypass the hooks, it's possible to do with `git commit --no-verify`.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,1 @@
+cargo fmt && cargo fix --allow-dirty --allow-staged
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,1 +1,1 @@
-cargo fmt && cargo fix --allow-dirty --allow-staged
+cargo fmt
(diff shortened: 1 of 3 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1 @@
+cargo fmt && cargo fix --allow-dirty --allow-staged
```
MaciejBaj, line 1: cargo fmt ok, cargo fix takes too long to be run on each commit

MaciejBaj (changes requested):

.githooks/pre-commit
palozano: Agree. Removed.

PR #599, opened 2023-01-25
chore: prep the shell runtime

AwesomeIbex:
I have been testing the shell runtime in a polkadot environment and modified the runtimes for anything we missed. Largely there was not a lot missing.

I've also removed the duplicate async environments we had. The general rule of thumb for async runtimes is to pick one or the other, NOT both. This probably made the node binary pretty big.

Also cleaned up some stuff that was left there until such a time we need it.

Tested:
- balances
- ran all day to see if any stoppages

Largely the shell is pretty simple


node/t3rn-parachain/src/chain_spec.rs
```diff
@@ -166,8 +205,9 @@ pub fn polkadot_config() -> ChainSpec {
         "t3rn_polkadot",
         ChainType::Live,
         move || {
+            // TODO: needs updating
             polkadot_genesis(
-                // Invulnerable collators
+                // Invulnerable collators FIXME: these are NOT the right collators
```
AwesomeIbex, line 210: We need to figure out the strategy on what we do here, maybe just pregen the keys and update this once we have. Then we can prefund the collators with the bond at genesis.

runtime/t3rn-parachain/src/lib.rs
```diff
@@ -38,24 +33,19 @@ use frame_support::{
...
 
-/// Executive: handles dispatch to the various modules.
-pub type Executive = frame_executive::Executive<
-    Runtime,
-    Block,
-    frame_system::ChainContext<Runtime>,
-    Runtime,
-    AllPalletsWithSystem,
->;
+pub const EXISTENTIAL_DEPOSIT: u128 = 500;
```
chiefbiiko, line 48: We could simplify this by setting it to 1. Would allow most users to forget about the existential deposit and probably only be relevant in situations where we do passive payouts. Cannot be zero though because it is used as divisor somewhere in the runtime.

MaciejBaj: Agreed

node/t3rn-parachain/src/chain_spec.rs
```diff
@@ -11,7 +10,12 @@ use sp_runtime::traits::{IdentifyAccount, Verify};
 use std::str::FromStr;
 use t3rn_primitives::monetary::TRN;
 
-const PARACHAIN_ID: u32 = 3000_u32;
+const PARACHAIN_ID: u32 = 3000;
+const SUPPLY: u128 = (TRN as u128) * 100_000_000; // 100 million TRN
+const CANDIDACY_BOND: u128 = (TRN as u128) * 10_000; // 10K TRN
+const DESIRED_CANDIDATES: u32 = 32;
+const SUDO: &str = "1t3rnvVous5FTJdqrxR5AQh7UGycPHk98rU63METkgPxbFE";
```
MaciejBaj, line 17:
```suggestion
const SUDO: &str = "t3UH3gWsemHbtan74rWKJsWc8BXyYKoteMdS78PMYeywzRLBX";
```

node/t3rn-parachain/src/chain_spec.rs
```diff
@@ -11,7 +10,12 @@ use sp_runtime::traits::{IdentifyAccount, Verify};
 use std::str::FromStr;
 use t3rn_primitives::monetary::TRN;
 
-const PARACHAIN_ID: u32 = 3000_u32;
+const PARACHAIN_ID: u32 = 3000;
```
MaciejBaj, line 13:
```suggestion
const PARACHAIN_ID: u32 = 3333;
```

node/t3rn-parachain/src/chain_spec.rs
```diff
@@ -186,12 +227,12 @@ pub fn polkadot_config() -> ChainSpec {
                 ],
                 // Prefunded accounts
                 vec![
-                    // Genesis Account: t3UH41t54cqKW72s9KSepTr1xW1r9F91aPfJvDfDgQf1CAgRC
-                    hex!("00a796547c96dcb02d365e3c0965ac0604575fa8cb0c0d98bd0e04e5e786db4d").into(),
+                    // Genesis Account: SUDO
```
MaciejBaj, line 230:
```suggestion
                    // Genesis Account: SUDO (t3UH3gWsemHbtan74rWKJsWc8BXyYKoteMdS78PMYeywzRLBX = hex!("0x00a6769855d6df941f09e0743f8879f66bad2dde6534a268dfe478449a16312b").into()
```

node/t3rn-parachain/src/chain_spec.rs
```diff
@@ -166,8 +206,9 @@ pub fn polkadot_config() -> ChainSpec {
         "t3rn_polkadot",
         ChainType::Live,
         move || {
+            // TODO: needs updating
             polkadot_genesis(
-                // Invulnerable collators
+                // Invulnerable collators FIXME: these are NOT the right collators
                 vec![
                     (
                         // Collator 1: t3W7yG2pkGdLogoX6KJm5KtPMMWBQygvcZArcjtjo5AsJPad2
```
MaciejBaj, line 214:
```suggestion
                        // Collator 1: "t3XXX7FGKAsG3pwE188CP91zCgt4p2mEQkdeELwocRJ4kCrSw"
```

MaciejBaj (changes requested):

MaciejBaj (approved):

PR #626, opened 2023-02-06
tests: retrofit zombienet to t0rn

AwesomeIbex:
This uses predefined releases to test with, speeding up zombienet much more.
Also adds some better testability to the scripts for downloading.

Tested here https://github.com/t3rn/t3rn/actions/runs/4102683866/jobs/7076121595


.github/workflows/release-t0rn.yml
```diff
@@ -2,6 +2,7 @@ name: t0rn Rococo Release
 
 on:
   push:
+    # This is a GLOB not a regex. Don't add escape sequences to me
```
3h4x, line 5: Double comment

MaciejBaj (approved):

PR #644, opened 2023-02-10
fix(t0rn): add slow adjusting fee 

AwesomeIbex:


MaciejBaj (changes requested):
+ mock + standalone + t3rn

AwesomeIbex:
> * mock + standalone + t3rn

I wouldnt do it for mock because then all determinism is removed from balance checks.

PR #733, opened 2023-03-14
fix(clock): round info config should be every 400 blocks

AwesomeIbex:
Updates round info to be every 400 blocks. This fixes NewRound spam.

### DIFF
--- a/pallets/clock/src/lib.rs
+++ b/pallets/clock/src/lib.rs
@@ -136,13 +136,17 @@ pub mod pallet {
                 head: n,
                 term,
             };
-            <CurrentRound<T>>::put(new_round);
-            Self::deposit_event(Event::NewRound {
-                index: new_round.index,
-                head: new_round.head,
-                term: new_round.term,
-            });
-            T::DbWeight::get().reads_writes(KILL_READS, KILL_WRITES)
+            if new_round.should_update(n) {
+                <CurrentRound<T>>::put(new_round);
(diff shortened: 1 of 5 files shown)


MaciejBaj (changes requested):
should also work, but RoundInfo in primitives/src/common.rs should be updated/fixed as well. 

right now: 

- term is set to 20 blocks - change to 400
    fn default() -> RoundInfo<B> {
        RoundInfo::new(1u32, 1u32.into(), 20u32.into())
    }
    
- should_update let's triggers "update" each block, so ignores the term requirements 
    /// Check if the round should be updated
    pub fn should_update(&self, now: B) -> bool {
        now - self.head >= self.term
    }

petscheit (approved):

pallets/clock/src/lib.rs
```diff
@@ -136,13 +136,17 @@ pub mod pallet {
...
                 term,
             };
-            <CurrentRound<T>>::put(new_round);
-            Self::deposit_event(Event::NewRound {
-                index: new_round.index,
-                head: new_round.head,
-                term: new_round.term,
-            });
-            T::DbWeight::get().reads_writes(KILL_READS, KILL_WRITES)
+            if new_round.should_update(n) {
```
MaciejBaj, line 139: 🥇

MaciejBaj (approved):

PR #739, opened 2023-03-15
feat(API)!: add execution_layer and parent_gateway_id to XDNS entry

petscheit:
# Summary of changes
1. Adding two new fields to XDNS that are helpful for client side applications. 

Execution Layer: 
- Differentiate between the different execution layers, e.g. Substrate & EVM
- For example, the client side logic for doing ths is the same for Ethereum & BSC

Parent Gateway:
- Maps gateway to its parent chain. e.g. para -> relay
- Needed for doing child inclusion proofs

I have also removed the xdns genesis records, as they are not used in any case.

## Acceptance Checklist

- [x] New feature (non-breaking change which adds functionality)
    - [x] Have you **seen it working** in a development environment?
    - [x] Have you **written tests** for **happy** and **unhappy** paths?
    - [x] Have you implemented this feature as per the **issue acceptance criteria**?
- [x] Breaking change (a feature that would cause existing functionality not to work as expected
    - [ ] Is the breaking change **documented**?
- [ ] CI/CD
- [ ] Documentation Update
- [ ] Substrate
    - [ ] RPC methods?
    - [ ] Runtime versioning?
    - [ ] Benchmarks?

## Further comments

### DIFF
--- a/client/packages/cli/commands/register/substrate.ts
+++ b/client/packages/cli/commands/register/substrate.ts
@@ -13,7 +13,8 @@ export const registerSubstrate = async (circuit: ApiPromise, gatewayData: any, e
         provider: new WsProvider(gatewayData.rpc),
     })
 
-    if(gatewayData.registrationData.parachain === null) { // relaychain
+
+    if(gatewayData.registrationData.parentChainId === null) { // relaychain
         return registerRelaychain(circuit, target, gatewayData, epochsAgo)
     } else {
         return registerParachain(circuit, target, gatewayData)
@@ -32,6 +33,8 @@ const registerRelaychain = async (circuit: ApiPromise, target: ApiPromise, gatew
         gateway_genesis: await createGatewayGenesis(circuit, target),
         gateway_sys_props: createGatewaySysProps(circuit, gatewayData.registrationData.gatewaySysProps),
         allowed_side_effects: circuit.createType('Vec<AllowedSideEffect>', gatewayData.registrationData.allowedSideEffects),
+        parent_gateway_id: circuit.createType('Option<ChainId>', gatewayData.registrationData.parentChainId),
+        execution_layer: circuit.createType('ExecutionLayer', gatewayData.registrationData.executionLayer),
         registration_data: circuit.createType('GrandpaRegistrationData', [
             registrationHeader.toHex(),
             Array.from(authorities),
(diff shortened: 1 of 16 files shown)


client/packages/cli/commands/register/substrate.ts
```diff
@@ -32,6 +33,8 @@ const registerRelaychain = async (circuit: ApiPromise, target: ApiPromise, gatew
         gateway_genesis: await createGatewayGenesis(circuit, target),
         gateway_sys_props: createGatewaySysProps(circuit, gatewayData.registrationData.gatewaySysProps),
         allowed_side_effects: circuit.createType('Vec<AllowedSideEffect>', gatewayData.registrationData.allowedSideEffects),
+        parent_gateway_id: circuit.createType('Option<ChainId>', gatewayData.registrationData.parentChainId),
```
MaciejBaj, line 36: How about 3 separate Gateway entries to differentiate between relays/beacons, shards/parachains, token on specific beacon/shard?

```
fn register_beacon_gateway(
    origin: OriginFor<T>,
    gateway_id: ChainId,
    codec: t3rn_abi::recode::Codec,
    gateway_abi_specs: GatewayABISpecs,
    gateway_genesis: GatewayGenesisConfig,
    token_props: TokenSysProps,
    allowed_standard_sfx_abi: Vec<Sfx4bId>,
    security_portal: u8,
    escrow_address: Option<Bytes>,
) {
}

fn register_shard_gateway(
    origin: OriginFor<T>,
    beacon_id: ChainId,
    gateway_id: ChainId,
    gateway_abi_specs: GatewayABISpecs,
    escrow_address: Option<Bytes>,
    token_props: TokenSysProps,
    allowed_standard_sfx_abi: Vec<Sfx4bId>,
) {
}

// Add a new token to the registry.
// Derives all settings from the gateway.
fn add_token(
    origin: OriginFor<T>,
    gateway_id: ChainId,
    token_base: ChainId,
    token_props: TokenSysProps,
) 
```

MaciejBaj (changes requested):

client/packages/cli/commands/register/substrate.ts
MaciejBaj: security_portal: u8 is a reference to Instance of Light Client we registered in t3rn

client/packages/cli/commands/register/substrate.ts
```diff
@@ -32,6 +33,8 @@ const registerRelaychain = async (circuit: ApiPromise, target: ApiPromise, gatew
...
+        execution_layer: circuit.createType('ExecutionLayer', gatewayData.registrationData.executionLayer),
```
MaciejBaj, line 37: Prefer if u take ExecutionLayer from PR#667 - t3rn_abi::codec (either Rlp / Scale for now)

petscheit: Are we sure those will always stay this way? I need something like this in executor to map gateways to correct tx/proof generation logic

client/packages/cli/commands/register/substrate.ts
petscheit: 1) But these would all result in XDNS entries also right? Personally I would like to keep a generic extrinsic for this, but I agree that what we currently have isn't there yet.
2) What do we want to register token for? For reward payment asset? 
3) Why do you want to get rid of GatewayVendor?

Issue #716, opened 2023-03-13
Executor: Deamonize Application


petscheit:
### Item Type

Task

### Description

The executor needs to become a daemonized application for mainnet. This enables a number of important features.

- [ ] Shutdown Sequence:
In a production env, we need to be able to initiate a shutdown of the executor. This will cause the executor to not accept new executions, but finish the ones it's currently participating in. This will allow safe shutdowns, ensuring the executor is not slashed accidentally.

- [ ] Update Settings:
The executor is setup via a config file, specifying everything from strategy, to bidding to price sources. We want to be able to update these settings at runtime, ensuring that all settings are formatted correctly. The updated settings should then be persisted somehow.

### Acceptance Criteria

- As an Executor, I want to be able to shut down at any moment safely. I can be sure that no new execution is added, while my ongoing executions are finalized.
- As an Executor, I want to be able to update my configuration at runtime. For example, I want to stop accepting executions for a specific gateway that I previously served.

### What are the points assigned to this item?

None

### Is the issue ready?

- [X] This issue is ready

PR #781, opened 2023-04-04
refactor: daemonize executor

chiefbiiko:
# Summary of changes

Refactors the configuration flow of an executor. Now, multiple separately configured instances can be spun up on a single machine. Furthermore, the executor now allows graceful shutdowns which can be initiated by sending CTRL+C, upon which all pending executions are completed, and then the process exits.

#716

## Acceptance Checklist

- [x] New feature (non-breaking change which adds functionality)
    - [x] Have you **seen it working** in a development environment?
    - [ ] Have you **written tests** for **happy** and **unhappy** paths?
    - [x] Have you implemented this feature as per the **issue acceptance criteria**?
- [x] Breaking change (a feature that would cause existing functionality not to work as expected
    - [x] Is the breaking change **documented**?
- [ ] CI/CD
- [ ] Documentation Update
- [ ] Substrate
    - [ ] RPC methods?
    - [ ] Runtime versioning?
    - [ ] Benchmarks?

## Further comments

The config auto reloading during runtime will be implemented as part of the new state persistence mechanism.


client/packages/executor/src/index.ts
```diff
@@ -42,21 +49,94 @@ class InstanceManager {
...
         this.executionManager = new ExecutionManager(this.circuitClient, this.sdk, logger)
-        await this.executionManager.setup(config.gateways, config.vendors)
+        await this.executionManager.setup(config.gateways, []) //TODO 2nd arg vendors
+
+        this.registerExitListener()
 
         logger.info("Executor: setup complete")
     }
+
+    async loadConfig(name: string) {
```
palozano, line 75: Very nice

client/packages/executor/src/executionManager/index.ts
```diff
@@ -102,6 +102,31 @@ export class ExecutionManager {
         this.addLog({ msg: "Setup Successful" })
     }
 
+    /** Initiates the shutdown sequence. */
+    async shutdown() {
```
palozano, line 106: This is very good

palozano:
Currently, one of the tests wasn't passing locally and in the GH checks.

The test is `executionManager.ts > SideEffect > should roundtrip sfx`.

In said test, there's a call to `SideEffect.fromJSON()`, where a `new Gateway` is created (line 249 of `client/packages/executor/src/executionManager/sideEffect.ts`).

The value that is passed to the Gateway is `o.misc.xdns`, which only holds this data:

```json
o.misc.xdns -> { id: 'ROCO', createSfx: { tran: [Function: tran] } }
```

Therefore, in the Gateway constructor, when `this.id = record.gateway_record.gateway_id.toHuman();` is called (record is o.misc.xdns), the test fails, because it cannot find it.

The complete XDNS record should be passed to the Gateway constructor call (`T3rnPrimitivesXdnsFullGatewayRecord`).

client/packages/executor/src/executionManager/sideEffect.ts
```diff
@@ -178,10 +206,57 @@ export class SideEffect extends EventEmitter {
...
+                    return o.sideEffect.action.toString()
+                },
+            },
+            target: {
+                toU8a() {
+                    return Uint8Array.from(Buffer.from(o.misc.gatewayId))
+                },
+            },
         }
+        sideEffect = new Gateway(o.misc.xdns).createSfx[o.sideEffect.action](sideEffect) as T3rnTypesSideEffect
```
petscheit, line 249: We can use the sdk instance here directly, so `sdk.gateways[gatewayId].createSfx[o.sideEffect.action](sideEffect)`

chiefbiiko: The problem is that initializing the sdk (i.e. initializing its gateways) is not feasible at this point because we are in the prescribed sync fromJSON, whose signature we do not control, whereas `sdk.init` is async.

chiefbiiko:
moved to a new branch [new-x716](https://github.com/t3rn/t3rn/tree/new-x716)

PR #753, opened 2023-03-22
refactor: adapt Portal, Grandpa FV and XDNS to new LightClient trait

MaciejBaj:
# Summary of changes

**Summarize** the description of your changes. 

If it fixes a bug or creates a feature, link the issue.

## Acceptance Checklist

What types of changes does your code introduce?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue):
    - [ ] Have you **written tests **to protect against future occurrences of the bug?
- [ ] New feature (non-breaking change which adds functionality)
    - [ ] Have you **seen it working** in a development environment?
    - [ ] Have you **written tests** for **happy** and **unhappy** paths?
    - [ ] Have you implemented this feature as per the **issue acceptance criteria**?
- [ ] Breaking change (a feature that would cause existing functionality not to work as expected
    - [ ] Is the breaking change **documented**?
    - [ ] Has any previous changes been **deprecated**?
- [ ] CI/CD
    - [ ] If introducing new actions, have you **looked at the action code** for anything spurious?
    - [ ] Have you written **custom scripts** for things that could be actions already on the marketplace?
    - [ ] If introducing new actions, have you **pegged a static version**?
- [ ] Documentation Update
    - [ ] Have you checked other documentation, such as the docs repository?
    - [ ] If updating script documentation, have you **tested it on both environments**?
- [ ] Substrate
    - [ ] RPC methods?
    - [ ] Chainspec, both JSON specs & the module?
    - [ ] Runtime versioning?
    - [ ] Benchmarks?
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

Feel free to explain in further detail your choices if this is a significant change.

PR #786, opened 2023-04-05
chore: remove unused dependencies and update token type for eth2 integration

petscheit:
# Summary of changes

**Summarize** the description of your changes. 

## Acceptance Checklist
- removes a bunch of unused code
- wraps the token info in XDNS with an enum in preparation for eth2 integration
- adds execution vendor field to XDNS

- [ ] New feature (non-breaking change which adds functionality)
    - [ ] Have you **seen it working** in a development environment?
    - [ ] Have you **written tests** for **happy** and **unhappy** paths?
    - [ ] Have you implemented this feature as per the **issue acceptance criteria**?
- [ ] Breaking change (a feature that would cause existing functionality not to work as expected
    - [ ] Is the breaking change **documented**?
- [ ] CI/CD
- [ ] Documentation Update
- [ ] Substrate
    - [ ] RPC methods?
    - [ ] Runtime versioning?
    - [ ] Benchmarks?

## Further comments


MaciejBaj (dismissed):

client/packages/cli/config/setup.ts
```diff
@@ -40,11 +44,39 @@ export default {
...
+                        id: 42
+                    }
+                },
+                allowedSideEffects: [["tran", 4]],
+            }
+        },
+        {
+            name: "Ethereum",
+            id: "eth2",
+            token_id: "eth2",
```
AwesomeIbex, line 63:
```suggestion
            tokenId: "eth2",            
```

Assuming camelCase here too.

node/standalone/Cargo.toml
```diff
@@ -22,7 +22,6 @@ path = "src/main.rs"
 async-std              = { version = "1.10.0" }
```
AwesomeIbex, line 22: please remove this too

node/t0rn-parachain/Cargo.toml
```diff
@@ -34,7 +34,6 @@ pallet-3vm-evm-rpc       = { path = "../../pallets/evm/rpc", package = "pallet-e
 # Extras
 async-std              = { version = "1.10.0" }
```
AwesomeIbex, line 35: same here

pallets/xdns/src/lib.rs
```diff
@@ -403,20 +409,29 @@ pub mod pallet {
         fn add_new_token(
             token_id: [u8; 4],
             gateway_id: [u8; 4],
-            token_props: TokenSysProps,
+            token_props: TokenInfo,
         ) -> DispatchResult {
             // early exit if record already exists in storage
             if <Tokens<T>>::contains_key(token_id) {
```
AwesomeIbex, line 415: how do we not know that this collides with pallet-assets/asset-registry? How would we use it?

primitives/src/lib.rs
```diff
@@ -138,112 +155,126 @@ pub struct GatewayGenesisConfig {
...
-    pub token_decimals: u8,
+pub enum TokenInfo {
+    Substrate(SubstrateToken),
+    Ethereum(EthereumToken),
 }
 
-impl TryFrom<&ChainId> for TokenSysProps {
+#[derive(Clone, Eq, PartialEq, Encode, Decode, Debug, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct SubstrateToken {
```
AwesomeIbex, line 167: tbh im certain something like this already exists

AwesomeIbex (changes requested):
A lot of duplicate functionality in XDNS here, do we want to use that for assets now?

pallets/xdns/src/lib.rs
petscheit: This is used for whitelisting assets for SFX invocations for specific gateways. I honestly don't really know how this should tie into pallet-assets.

petscheit:
I made these asset changes only to support evm assets, which require different fields. The asset functionality was added with the latest XDNS refactor, so really no good idea of how that ties into xbi

pallets/xdns/src/lib.rs
AwesomeIbex: Pallet assets has a metadata field here, so we'd ensure that the token id exists for this. Otherwise we will have collisions when we come to use the assets registered here, and have to double the lookups when using assets already onboarded in the system.

For these gateway assets, we'd just register them with 0 ED, mint no tokens, the owner is sudo & the metadata is just this.

AwesomeIbex:
> I made these asset changes only to support evm assets, which require different fields. The asset functionality was added with the latest XDNS refactor, so really no good idea of how that ties into xbi

It's purely an architectural issue. XDNS is about registering information about a target. If a target has assets then they live in our thing that knows about assets. I understand that this PR didn't implement the functionality so I'll make a ticket to fix the issue.

petscheit:
#753 introduced this, I merely extended to support evm assets.

AwesomeIbex (dismissed):

PR #918, opened 2023-05-02
feat: implement inflation rewards to attesters and executors

MaciejBaj:
# Summary of changes

**_Summarize_** the description of your changes.

# Acceptance Checklist

## Dependencies

- [ ] Any teams which manage the dependencies have been notified of the breaking changes

## Bug (non-breaking change which fixes an issue):

- [ ] Have you **_written tests_** to protect against future occurrences of the bug?

## Feature (non-breaking change which adds functionality)

- [ ] Have you **_seen it working_** in a development environment?
- [ ] Have you **_written tests_** for **_happy_** and **_unhappy_** paths?
- [ ] Have you implemented this feature as per the **_issue acceptance criteria_**?

### Substrate

If this change involves substrate, have you remembered:

- [ ] Storage Migration?
- [ ] RPC methods?
- [ ] Runtime versioning?
- [ ] Benchmarks?

## Breaking change (a feature that would cause existing functionality not to work as expected)

- [ ] Is the breaking change **_documented_**?
- [ ] Are your commits fully representative of the change?
- [ ] Has any previous functionality been **_deprecated_** and versioned?

## CI/CD

- [ ] If introducing new actions, have you **_looked at the action code_** for anything spurious?
- [ ] Have you written **_custom scripts_** for things that could be actions already on the marketplace?
- [ ] If introducing new actions, have you **_pegged a static version_**?

## Documentation Update

- [ ] If updating script documentation, have you **_tested it on both environments_**?

## Further comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v1.21.0-rc.0] - 2023-05-12
+### :sparkles: New Features
+- [`6b24f01`](https://github.com/t3rn/t3rn/commit/6b24f019dde78e52301b5cad11f2fa74ea1287ab) - activate portal rpc service in t0rn runtime *(PR [#968](https://github.com/t3rn/t3rn/pull/968) by [@petscheit](https://github.com/petscheit))*
+
+
 ## [v1.20.1-rc.0] - 2023-05-12
 ### :bug: Bug Fixes
 - [`9367cb7`](https://github.com/t3rn/t3rn/commit/9367cb71c174d4f18e40493f5092ed0ff02d6e1d) - expose portal rpc endpoint in t0rn runtime *(PR [#967](https://github.com/t3rn/t3rn/pull/967) by [@petscheit](https://github.com/petscheit))*
@@ -1107,3 +1112,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [v1.19.2-rc.0]: https://github.com/t3rn/t3rn/compare/v1.19.1-rc.0...v1.19.2-rc.0
 [v1.20.0-rc.0]: https://github.com/t3rn/t3rn/compare/v1.19.2-rc.0...v1.20.0-rc.0
 [v1.20.1-rc.0]: https://github.com/t3rn/t3rn/compare/v1.20.0-rc.0...v1.20.1-rc.0
+[v1.21.0-rc.0]: https://github.com/t3rn/t3rn/compare/v1.20.1-rc.0...v1.21.0-rc.0
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12445,7 +12445,7 @@ dependencies = [
 
 [[package]]
 name = "t0rn-parachain-collator"
-version = "1.20.1-rc.0"
+version = "1.21.0-rc.0"
 dependencies = [
  "clap",
  "cumulus-client-cli",
@@ -12469,6 +12469,7 @@ dependencies = [
  "log",
  "pallet-contracts-rpc",
  "pallet-evm-rpc",
+ "pallet-portal-rpc",
  "pallet-transaction-payment-rpc",
  "pallet-xdns",
  "pallet-xdns-rpc",
@@ -12520,7 +12521,7 @@ dependencies = [
 
 [[package]]
 name = "t0rn-parachain-runtime"
-version = "1.20.1-rc.0"
+version = "1.21.0-rc.0"
 dependencies = [
  "circuit-runtime-types",
  "cumulus-pallet-aura-ext",
--- a/docs/main/docs/collator/testnet/testnet-collator.md
+++ b/docs/main/docs/collator/testnet/testnet-collator.md
@@ -5,7 +5,7 @@ sidebar_position: 1
 # Run a t0rn Testnet Collator
 
 
-This guide outlines the essential minimum of steps required to run a collator for t0rn - a release candidate of t3rn on the Rococo testnet. This guide uses the `v1.20.1-rc.0` release, however always use the latest available version.
+This guide outlines the essential minimum of steps required to run a collator for t0rn - a release candidate of t3rn on the Rococo testnet. This guide uses the `v1.21.0-rc.0` release, however always use the latest available version.
 
 Make sure to have your machine setup for [Rust and Substrate development](https://docs.substrate.io/v3/getting-started/installation/).
 
@@ -39,7 +39,7 @@ We maintain collator binaries which we release alongside every runtime release. 
 
 ```sh
 curl -sSfL \
-  https://github.com/t3rn/t3rn/releases/download/v1.20.1-rc.0/t0rn-collator-v1.20.1-rc.0-x86_64-unknown-linux-gnu.gz \
+  https://github.com/t3rn/t3rn/releases/download/v1.21.0-rc.0/t0rn-collator-v1.21.0-rc.0-x86_64-unknown-linux-gnu.gz \
 | gunzip > ~/t0rn/circuit-collator
 ```
 
@@ -56,17 +56,17 @@ To associate your node to the correct network we need to provide the t0rn chain 
 ```sh
 curl -sSfL \
   -o ~/t0rn/specs/rococo.raw.json \
-  https://raw.githubusercontent.com/t3rn/t3rn/v1.20.1-rc.0/specs/rococo.raw.json
+  https://raw.githubusercontent.com/t3rn/t3rn/v1.21.0-rc.0/specs/rococo.raw.json
 
 curl -sSfL \
   -o ~/t0rn/specs/t0rn.raw.json \
-  https://raw.githubusercontent.com/t3rn/t3rn/v1.20.1-rc.0/specs/t0rn.raw.json
+  https://raw.githubusercontent.com/t3rn/t3rn/v1.21.0-rc.0/specs/t0rn.raw.json
 ```
 
 ## Option 2: Pull latest Docker image
 
 ```sh
-docker pull ghcr.io/t3rn/t0rn-collator:v1.20.1-rc.0
+docker pull ghcr.io/t3rn/t0rn-collator:v1.21.0-rc.0
 ```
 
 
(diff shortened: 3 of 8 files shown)


pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,959 @@
...
+
+        /// The treasury's portion of the total inflation, expressed as a Perbill.
+        ///
+        /// Default: 2% (20 / 100)
+        #[pallet::constant]
+        type TreasuryInflation: Get<Perbill>;
+
+        /// The number of blocks in one year.
+        ///
+        /// Default: 5_256_000 (assuming 6s block time)
```
petscheit, line 74: We set up t3rn with 12 secs iirc

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,959 @@
...
+        }
+
+        pub fn distribute_collator_rewards(current_distribution: BalanceOf<T>) -> BalanceOf<T> {
+            let total_collators = Collators::<T>::iter().count() as u32;
+
+            if total_collators == 0 {
+                return Zero::zero()
+            }
+
+            let reward_per_collator = current_distribution / BalanceOf::<T>::from(total_collators);
```
petscheit, line 400: Let us reconsider the way we reward collators here. 

Looking at the collator rewards, IMO there are two main things to consider and ensure:
1. For the collators we want to optimize for raw collator count. The more, the better.
2. We want to make sure, collators and attestors are not competing, resulting in stake leaving attestors and moving to collators. The network doesn't really benefit from more stake in collators as its dead/non-slashable staking


With these two things in mind, I feel the current setup is not optimal. In the end, the APY between collators and attestors should balance to be roughly the same (considering hardware requirements for collators).

The reward factor difference between attestors and collators is x2.2 (1.1% vs 0.5%). Considering the hardware costs, let's roughly say 30% of the stake will be with collators. I feel like we are wasting stake that can be rather with attesters to secure the network. The stake of an attester doesn't provide any benefit to the network, so we shouldn't reward it (especially as it's competing with attesters). Instead, we could consider requiring a min amount of stake, but don't reward anything above. This would result in excess stake going to attesters, securing the network.

In other news:
I still think the share of collator rewards should be higher, given that it economically secures the value of all cross-chain txs, in the worst case, over an entire slashing period.

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,1232 @@
...
+                if let Some(index) = attesters.iter().position(|(a, _n)| a == &attester) {
+                    // Update the existing nomination amount
+                    attesters[index] = (attester.clone(), total_nomination);
+                } else {
+                    // Add the new attester to the list
+                    attesters.push((attester.clone(), total_nomination));
+                }
+
+                // Sort the attesters by their nomination amount
+                attesters.sort_by(|(_a1, n1), (_a2, n2)| n2.cmp(n1));
```
petscheit, line 579: Shouldn't this only happen on shuffle?

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,959 @@
...
+                // Remove the accumulated settlement from the storage
+                AccumulatedSettlements::<T>::remove(&executor);
+
+                distibuted_rewards += capped_reward;
+            }
+
+            distibuted_rewards
+        }
+
+        pub fn process_accumulated_settlements() -> Weight {
```
petscheit, line 466: Not really a fan of this, as it's inflexible, a complex call, and rewards incomplete. I see the benefit of having it claimable in an on-chain manner, but also a lot of downsides, when it comes to different things we might want to reward (tx value for all sfx, speed, liq on target, etc). Always requires us to add more complexity and strain on the node. Maybe we could think of a hybrid approach that allows us to make it claimable, but also more flexible

petscheit (approved):

pallets/rewards/src/lib.rs
```diff
@@ -0,0 +1,959 @@
...
+            if IsDistributionHalted::<T>::get() {
+                return
+            }
+            // Calculate the available balance for distribution in the current period
+            let total_issuance = T::Currency::total_issuance();
+            let distribution_period = T::InflationDistributionPeriod::get();
+            let one_year_blocks = T::OneYear::get();
+            // Include TotalInflation in the calculation
+            let total_inflation = T::TotalInflation::get();
+            let inflated_total_issuance: BalanceOf<T> = total_inflation.mul_ceil(total_issuance);
```
chiefbiiko, line 262: Given the shell setup this "round" inflation seems off because within shell we already have 100M circulating (sudo). Think the `total_issuance` factor would need to be altered here.

runtime/t0rn-parachain/src/circuit_config.rs
```diff
@@ -125,6 +126,68 @@ impl pallet_clock::traits::OnHookQueues<Runtime> for GlobalOnInitQueues {
...
+    type Currency = Balances;
+    type DefaultCommission = DefaultCommission;
+    type Event = Event;
+    type MaxBatchSize = ConstU32<128>;
+    type MinAttesterBond = MinAttesterBond;
+    type MinNominatorBond = MinNominatorBond;
+    type RandomnessSource = RandomnessCollectiveFlip;
+    type RewardMultiplier = RewardMultiplier;
+    type ShufflingFrequency = ConstU32<400>;
+    type SlashAccount = EscrowAccount;
```
chiefbiiko, line 152: Can't we just use `Treasury::account_id()` here? Meaning that slashed funds get dumped back into the treasury.

runtime/t0rn-parachain/src/circuit_config.rs
```diff
@@ -125,6 +126,68 @@ impl pallet_clock::traits::OnHookQueues<Runtime> for GlobalOnInitQueues {
...
+}
+
+use t3rn_primitives::monetary::TRN;
+
+parameter_types! {
+    pub const TotalInflation: Perbill = Perbill::from_parts(4_400_0000); // 4.4%
+    pub const AttesterInflation: Perbill = Perbill::from_parts(1_100_0000); // 1.1%
+    pub const ExecutorInflation: Perbill = Perbill::from_parts(0_800_0000); // 0.8%
+    pub const CollatorInflation: Perbill = Perbill::from_parts(0_500_0000); // 0.5%
+    pub const TreasuryInflation: Perbill = Perbill::from_parts(2_000_0000); // 2%
```
chiefbiiko, line 162: Why inflate the treasury? Thought treasury funds would simply be allocated at genesis.

runtime/t0rn-parachain/src/circuit_config.rs
```diff
@@ -125,6 +126,68 @@ impl pallet_clock::traits::OnHookQueues<Runtime> for GlobalOnInitQueues {
...
+    type CollatorBootstrapRewards = CollatorBootstrapRewards;
+    type CollatorInflation = CollatorInflation;
+    type Currency = Balances;
+    type Event = Event;
+    type ExecutorBootstrapRewards = ExecutorBootstrapRewards;
+    type ExecutorInflation = ExecutorInflation;
+    type InflationDistributionPeriod = InflationDistributionPeriod;
+    type OneYear = OneYear;
+    type TotalInflation = TotalInflation;
+    type TreasuryAccount = EscrowAccount;
```
chiefbiiko, line 187: Also `Treasury::account_id()` here?

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,1232 @@
...
+                attestations: Vec::new(),
+                target: [0u8; 4],
+                created: Zero::zero(),
+                status: BatchStatus::Pending,
+            }
+        }
+    }
+
+    impl<Attester, Signature, BlockNumber> Batch<Attester, Signature, BlockNumber> {
+        pub fn new(target: [u8; 4], now: BlockNumber) -> Self {
```
chiefbiiko, line 80:
```suggestion
        pub fn new(target: TargetId, now: BlockNumber) -> Self {
```

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,1232 @@
...
+        CommitteeSizeTooLarge,
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::weight(10_000)]
+        pub fn register_attester(
+            origin: OriginFor<T>,
+            self_nominate_amount: BalanceOf<T>,
+            ecdsa_key: [u8; 33],
```
chiefbiiko, line 225: What's that extra byte (33) for ECDSA pubkeys?

pallets/rewards/src/lib.rs
MaciejBaj: process_accumulated_settlements is only adding the already-earned rewards for execution. 

distribute_executor_rewards adds the extra rewards on top of the executions. I see the point that extending those with extra rewards ruleset will enforce a runtime upgrade. 

we also have the bootstrap pool alloc. to apply the extra off-chain rules for distribution. I am up for keeping the simple subsidy of settlement_amount for executor for inflation as algorithmic rule at start

runtime/t0rn-parachain/src/circuit_config.rs
MaciejBaj: can change 👍

runtime/t0rn-parachain/src/circuit_config.rs
MaciejBaj: 2% per year for parachain treasury as per https://docs.t3rn.io/about_t3rn/token-economics/tokenomics-inflation

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,1232 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(feature = "runtime-benchmarks")]
+pub mod benchmarking;
+
+pub use crate::pallet::*;
+
+pub type TargetId = [u8; 4];
```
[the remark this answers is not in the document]
MaciejBaj: yeah, will re-use

pallets/attesters/src/lib.rs
MaciejBaj: https://wiki.hyperledger.org/display/BESU/SECP256R1+Support 

> For the secp256r1 curve, the private key is 256-bit integer (32 bytes) and the compressed public key is 257-bit integer (~ 33 bytes).

pallets/rewards/src/lib.rs
MaciejBaj: Typically it's a percentage increase in the total supply of the asset over time, so this holds up

pallets/rewards/src/lib.rs
MaciejBaj: will update the comments, the runtime config calculates for 12s:

>     pub const OneYear: BlockNumber = 2_628_000; // (365.25 * 24 * 60 * 60) / 12; assuming 12s block time
>     pub const InflationDistributionPeriod: BlockNumber = 100_800; // (14 * 24 * 60 * 60) / 12; assuming one distribution per two weeks
>     pub const TreasuryAccount: AccountId = AccountId::new([51u8; 32]);
>     pub const AvailableBootstrapSpenditure: Balance = 1_000_000 * (TRN as Balance); // 1 MLN UNIT

pallets/attesters/src/lib.rs
MaciejBaj: I'm updating sorted set after each nomination, but assigning new active set at shuffle

MaciejBaj: if there's no un-nominations the sorting isn't necessary therefore

petscheit (approved):

PR #923, opened 2023-05-03
feat: implement attesters checks for ecdsa signature scheme 

MaciejBaj:
# Summary of changes

**_Summarize_** the description of your changes.

# Acceptance Checklist

## Dependencies

- [ ] Any teams which manage the dependencies have been notified of the breaking changes

## Bug (non-breaking change which fixes an issue):

- [ ] Have you **_written tests_** to protect against future occurrences of the bug?

## Feature (non-breaking change which adds functionality)

- [x] Have you **_seen it working_** in a development environment?
- [x] Have you **_written tests_** for **_happy_** and **_unhappy_** paths?
- [ ] Have you implemented this feature as per the **_issue acceptance criteria_**?

### Substrate

If this change involves substrate, have you remembered:

- [ ] Storage Migration?
- [ ] RPC methods?
- [ ] Runtime versioning?
- [ ] Benchmarks?

## Breaking change (a feature that would cause existing functionality not to work as expected)

- [ ] Is the breaking change **_documented_**?
- [ ] Are your commits fully representative of the change?
- [ ] Has any previous functionality been **_deprecated_** and versioned?

## CI/CD

- [ ] If introducing new actions, have you **_looked at the action code_** for anything spurious?
- [ ] Have you written **_custom scripts_** for things that could be actions already on the marketplace?
- [ ] If introducing new actions, have you **_pegged a static version_**?

## Documentation Update

- [ ] If updating script documentation, have you **_tested it on both environments_**?

## Further comments


pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,864 @@
...
+        }
+
+        #[pallet::weight(10_000)]
+        pub fn nominate(
+            origin: OriginFor<T>,
+            attester: T::AccountId,
+            amount: BalanceOf<T>,
+        ) -> DispatchResult {
+            let nominator = ensure_signed(origin)?;
+            Self::do_nominate(nominator.clone(), attester.clone(), amount);
```
MaciejBaj, line 376:
```suggestion
            Self::do_nominate(nominator.clone(), attester.clone(), amount)?;
```

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,864 @@
...
+    #[pallet::generate_store(pub(super) trait Store)]
+    #[pallet::without_storage_info]
+    pub struct Pallet<T>(_);
+
+    #[pallet::storage]
+    #[pallet::getter(fn attesters)]
+    pub type Attesters<T: Config> = StorageMap<_, Blake2_128Concat, T::AccountId, AttesterInfo>;
+
+    #[pallet::storage]
+    pub type CurrentCommittee<T: Config> = StorageValue<_, Vec<T::AccountId>, ValueQuery>;
```
petscheit, line 127: We should consider doing this index based. This makes the signature verification on the target a lot cheaper, as we can get a pub key based on the index, instead of looping. This was done in Ethereum and we just need to sort the committee keys based on some condition when initializing and updating the committee

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,864 @@
...
+
+                    sfx_attestation.for_ = attestation_for;
+                    sfx_attestation.status = status;
+
+                    Attestations::<T>::insert(sfx_hash, sfx_attestation.clone());
+
+                    let mut target_batches = Batches::<T>::get(target).unwrap_or_default();
+                    let current_block = frame_system::Pallet::<T>::block_number();
+
+                    if target_batches.is_empty() {
```
petscheit, line 287: I think this batching logic can be made significantly more efficient by reducing the number of signature checks. IMO the elements that are being signed are queued here, once the queue is full, we throw the elements(sfx, etc) in a merkle tree and let the attestors only sign the root. This way we only do one signature check per attestor and batch

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,864 @@
...
+                },
+            }
+
+            Self::deposit_event(Event::AttestationSubmitted(account_id));
+
+            Ok(())
+        }
+
+        #[pallet::weight(10_000)]
+        pub fn commit_batch(
```
petscheit, line 328: What is this supposed to do?

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,864 @@
...
+            Ok(())
+        }
+    }
+
+    impl<T: Config> Pallet<T> {
+        pub fn committee_size() -> usize {
+            T::CommitteeSize::get() as usize
+        }
+
+        pub fn do_nominate(
```
petscheit, line 387: Thoughts on pallet-staking from moonbeam? I looked through it, and to me it seems like it gives us what we need. Feel like there are a lot of things to mess up here.

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,864 @@
...
+
+            nomination.1 += amount;
+
+            let nomination_amount = nomination.1;
+
+            // Update the nomination storage item
+            Nominations::<T>::insert(&attester, nomination);
+
+            // Update the sorted list of nominated attesters
+            SortedNominatedAttesters::<T>::try_mutate(|attesters| {
```
petscheit, line 411: I think we should only trigger after a certain period is complete, in this form there are some major slashing implications

pallets/attesters/src/lib.rs
```diff
@@ -0,0 +1,864 @@
...
+
+            full_shuffle
+        }
+    }
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_initialize(n: T::BlockNumber) -> Weight {
+            // Check if a shuffling round has passed
+            if (n % T::ShufflingFrequency::get()).is_zero() {
```
petscheit, line 481: We should consider always assigning the new attestor set I shuffling period in advance, so that we always have current and next committee. And then we need to generate a proof, where the current committee signes of the next

petscheit (approved):
Good stuff!

MaciejBaj:
continued in #918

PR #1174, opened 2023-07-17
feat: implement bridge contract to receive remote orders

MaciejBaj:

PR #1205, opened 2023-07-24
test: cover xtx id calculations for nonces 0,1 and 2

MaciejBaj:
test: cover xtx id calculations for nonces 0,1 and 2

cover test cases to match the expected keccak hashes for nonces 0,1 and 2 at #1174 

### Summary by OpenAI

Release Notes:
- Test: Added a new test case to cover the calculation of expected values for nonce 0, 1, and 2 in `primitives/src/circuit/types.rs`. It verifies transaction IDs and side effect IDs against expected hash values.

> "In the realm of tests, a new case emerges,
> Covering nonces with different surges.
> Hash values verified, expectations met,
> Bugs uncovered, no room for regret."


3h4x (approved):
👍

PR #1221, opened 2023-07-27
fix: stop committee shuffle if previous transition request still awaits attestation

MaciejBaj:
## fix: stop committee shuffle if previous transition request still awaits attestation

committee shuffle on t0rn was still awaiting the pending attestations, while the committee shuffle continued. 

This PR adds the storage entry: CommitteeTransitionOn which marks batch index committee was shuffled on each attestation target. If before reaching next shuffling period the batch containing committee transition still awaits attestation was pending, do not shuffle committee. Doing otherwise leads to:
- attestations can be accepted from 2 different committees (previous and current) which is unlikely to be verified by attestation verifier contract. 

There's a new helper extrinsic that reads all pending attestations to all registered attestation targets and Emit via event: call::read_pending_batches()

### Summary by OpenAI

**New Feature**
- The `request_next_committee_attestation` function now returns a vector of tuples, enhancing the information returned.
- Added a new storage entry `CommitteeTransitionOn` to track committee transitions.
- Introduced a helper extrinsic for reading pending attestations, improving data accessibility.
- An event `call::read_pending_batches()` is now emitted, providing real-time updates on batch processing.

> 🎉 With tuples in vectors, we stride,
> In storage entries, secrets hide. 🗄️
> Extrinsic helpers pave the way,
> While events keep us in the fray. 🚀
> To better code, we've taken a ride! 🎊


pallets/attesters/src/lib.rs
```diff
@@ -1639,8 +1686,17 @@ pub mod pallet {
                 // Call shuffle_committee
                 Self::shuffle_committee();
                 aggregated_weight += T::DbWeight::get().reads_writes(2, 2);
-                Self::request_next_committee_attestation();
-                aggregated_weight += T::DbWeight::get().reads_writes(2, 2);
+
+                Self::deposit_event(Event::ShufflingCompleted(
```
3h4x, line 1690: Looks good!

3h4x (approved):
lgtm

PR #1234, opened 2023-08-09
feat: implement substrate price estimation

ahkohd:
### Summary by OpenAI

**New Feature**
- Added support for estimating gas fees and maximum rewards in a Substrate-based blockchain.
- Introduced new functions and types related to Substrate estimation.
- Refactored existing functions to accommodate the changes.

> 🎉 With Substrate now in sight, 
> Our SDK takes flight. 
> Estimating gas, rewards at max,
> On blockchain tracks, there's no turning back! 🚀


client/packages/sdk/src/price-estimation/__tests__/substrate.test.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+const buildSfx = (circuit: ApiPromise
+) => {
+  return {
+    sideEffects: createType("Vec<T3rnTypesSfxSideEffect>", [{
+      target: "roco",
+      maxReward: 1000000,
+      insurance: 1000000,
+      encodedArgs: ["0x0", "0x1"],
```
petscheit, line 36: I think its important to have proper arg length here, as tx bytes is part of the tx cost

client/packages/sdk/src/price-estimation/index.ts
```diff
@@ -152,13 +141,13 @@ export async function estimateBidAmount<T extends EstimateEthActionParams | Spee
  * @returns The maximum reward estimate
 */
 
-export async function estimateMaxReward<T extends EstimateEthActionParams | SpeedMode>({
-  action, asset: baseAsset, target, targetAmount, targetAsset, overSpendPercent = 0.5, args
+export async function estimateMaxReward<T extends EstimateParams>({
+  action, baseAsset, target, targetAmount, targetAsset, overSpendPercent = 0.5, args
 }: EstimateMaxReward<T>): Promise<MaxRewardEstimation> {
-  const targetNativeAsset = mapTargetToNativeAsset(target);
+  const targetNativeAsset = getTargetNativeAsset(target);
```
petscheit, line 147: you can get this via XDNS also

ahkohd: In our approach to price estimation within this module, I've opted to avoid RPC calls and keep everything localized. To that end, I've implemented a method `NativeAssetMap.setFor(target, symbol)`.

If you ever need to change the default asset symbols, you can do so by:

1. Making an XDNS call for the target you want to change.
2. Calling `NativeAssetMap.setFor(target, symbol)` with the updated symbol.

This way, we can keep everything within this module and have a more streamlined process.

3h4x (changes requested):
This should be integrated with CLI which now for estimate returns error
`✖ TypeError: Cannot destructure property 'tx' of 'undefined' as it is undefined.`

Also functional test should be in README as it requires node-standalone to run.

ahkohd:
> This should be integrated with CLI, which now for estimate returns error `✖ TypeError: Cannot destructure property 'tx' of 'undefined' as it is undefined.`
> 
> Also functional test should be in README as it requires node-standalone to run.

I'll raise a new PR for the issue once I have this PR merged.

MaciejBaj (approved):

PR #1260, opened 2023-08-23
fix(executor): balance calculation and new metrics

3h4x:
### Summary by OpenAI

**New Feature**
- Removed `executorBids` counter metric from the `Prometheus` class.
- Introduced a new metric, `gateway_height`, to the `Prometheus` class for tracking the height of a gateway.
- Updated the `prometheus.gatewayHeight` metric with vendor and height information in `executionManager`.
- Added a new metric for gateway height in the `SubstrateRelayer` class.
- Incremented a Prometheus counter to track executor execution completion.

> 🎉 With metrics refined, our path is aligned,
> Gateway's height we now can find. 
> No more blind, in data we bind,
> To the rhythm of code, our minds are signed. 🚀

### DIFF
--- a/client/packages/executor/src/utils/index.ts
+++ b/client/packages/executor/src/utils/index.ts
@@ -3,6 +3,7 @@ import { ApiPromise } from "@polkadot/api";
 import { u8aToHex } from "@polkadot/util";
 import { xxhashAsU8a } from "@polkadot/util-crypto";
 import { BN } from "@polkadot/util";
+import { Sdk } from "@t3rn/sdk";
 import { default as pino, Logger } from "pino";
 
 export async function getStorage(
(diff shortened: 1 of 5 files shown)


client/packages/executor/src/utils/index.ts
```diff
@@ -88,3 +89,21 @@ export function createLogger(name: string, logsDir?: string): Logger {
...
+
+export async function getBalanceWithDecimals(
+  client: Sdk["client"] | ApiPromise,
+  address: string,
+  decimals: number = 12,
+): Promise<number> {
+  const balance = await client.query.system.account(address);
+  const balanceBN = new BN(balance.data.free.toString());
+
+  // Assuming `balance.data.free` is the balance in the lowest unit (smallest decimal)
```
ahkohd, line 101: Nice!

ahkohd (approved):
LGTM!

PR #1291, opened 2023-09-08
test: check for drained ImportedHeaders at grandpa LC reset

MaciejBaj:
### Summary by OpenAI

**Test:**
- Added assertions to verify the behavior of `ImportedHeaders` and `ImportedHashes` data structures during a reset operation. This ensures that all expected keys are present in `ImportedHeaders` and `ImportedHashes` is empty post-reset.

> 🎉 Here's to the code that's now more robust, 🥂
> With tests for our data, in correctness we trust. 🛡️
> Reset operations, under the test's gaze, 👀
> Ensuring our structures, in order always stay. 🏗️


3h4x (approved):
very good! 🪨

PR #1295, opened 2023-09-12
feat: XCM Transfers - updated t0rn XCM config

chexware:


3h4x:
@chexware Can you create this PR from t3rn and not from forked repo. Otherwise GHA wont run correctly. 🙏

PR #1307, opened 2023-09-18
fix: check for above-zero target height increase before setting isActive

MaciejBaj:
### Summary by OpenAI

**New Features:**
- Added checks for above-zero target height increase in `pallets/xdns/src/lib.rs`.
- Introduced new test case `test_determines_none_for_two_non_increasing_elements` in `primitives/src/lib.rs`.
- Modified the `DisabledValidators` type in `runtime/t0rn-parachain/src/parachain_config.rs`.
- Updated import statements in `node/parachain/src/rpc.rs`, `node/standalone/src/rpc.rs`.
- Implemented new functions and methods related to side effects and transactions in `pallets/circuit/src/lib.rs`, `pallets/circuit/src/machine/test.rs`, `primitives/src/circuit/traits.rs`.
- Added a new function `fetch_all_active_xtx` to the `PortalRuntimeApi` trait and its implementations across multiple files.
- Removed a call to `XDNS::process_overview(System::block_number())` in `runtime/mini-mock/src/lib.rs`.
- Enabled serialization and deserialization for the `SideEffect` struct in `types/src/sfx.rs`.

> 🎉 Code changes are here, making things clear, 🚀
> New features abound, improvements all around. 🌟
> With each PR merge, we surge, 💪
> On the path of innovation, we continue our journey without reservation. 🛤️

### DIFF
--- a/runtime/t0rn-parachain/src/parachain_config.rs
+++ b/runtime/t0rn-parachain/src/parachain_config.rs
@@ -107,7 +107,7 @@ impl pallet_session::Config for Runtime {
 impl pallet_aura::Config for Runtime {
     type AllowMultipleBlocksPerSlot = ConstBool<false>;
     type AuthorityId = AuraId;
-    type DisabledValidators = ();
+    type DisabledValidators = Session;
     type MaxAuthorities = MaxAuthorities;
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7306,6 +7306,7 @@ dependencies = [
  "sp-core",
  "sp-rpc",
  "sp-runtime",
+ "t3rn-types",
 ]
 
 [[package]]
@@ -7316,7 +7317,9 @@ dependencies = [
  "parity-scale-codec",
  "sp-api",
  "sp-runtime",
+ "sp-std 8.0.0",
  "t3rn-primitives",
+ "t3rn-types",
 ]
 
 [[package]]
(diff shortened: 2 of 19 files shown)


runtime/t0rn-parachain/src/parachain_config.rs
```diff
@@ -107,7 +107,7 @@ impl pallet_session::Config for Runtime {
 impl pallet_aura::Config for Runtime {
     type AllowMultipleBlocksPerSlot = ConstBool<false>;
     type AuthorityId = AuraId;
-    type DisabledValidators = ();
+    type DisabledValidators = Session;
```
3h4x, line 110: > A way to check whether a given validator is disabled and should not be authoring blocks. Blocks authored by a disabled validator will lead to a panic as part of this module’s initialization.

https://docs.rs/pallet-aura/latest/pallet_aura/pallet/trait.Config.html#associatedtype.DisabledValidators

So what will happen when we will switch it to `Session`?

MaciejBaj: Session exposes interface to check disabled validators. currently however the list of disabledValidators is empty on t0rn so we'd need to pin slashing rules that would ask Session::disableIndex

gvko (changes requested):
minor stuff.
Also, build doesn't suceed.

gvko (approved):
build & test passing, all good

PR #1311, opened 2023-09-22
chore: disable inactive t0rn collators once per 6h session

MaciejBaj:
### Summary by OpenAI

**New Feature:**
- Added a new function `find_inactive_authors` in the `pallet` module to filter out inactive authors based on the current session validators.
- Introduced a new block number constant `SESSION_PERIOD` in the `hooks` module, representing a 6-hour session period.
- Implemented logic to disable inactive validators from the session if they haven't produced any blocks during the session period.

> 🎉 With each passing block, we keep track,
> Of validators who do their part or slack.
> Inactive ones, we now can find,
> Ensuring our chain's health, aligned. 🚀

### DIFF
--- a/pallets/rewards/src/lib.rs
+++ b/pallets/rewards/src/lib.rs
@@ -718,6 +718,16 @@ pub mod pallet {
             }
         }
 
+        pub fn find_inactive_authors(
+            current_session_validators: Vec<T::AccountId>,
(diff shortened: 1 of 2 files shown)


3h4x (approved):
let's go! 🚀

mail731212:
@zannis

PR #1404, opened 2023-10-27
fix(grandpa-ranger): sending limited ranges to not exceed blocksize

coun7zero:
### Summary by CodeRabbit

- Refactor: Updated the `batches_max` configuration value from 10 to 2 in the Grandpa Ranger module, optimizing the data processing by limiting the maximum length of the batches array.
- Style: Improved code readability by organizing nested objects and adding missing commas.
- Refactor: Removed an unnecessary slicing operation on the `range` array, enhancing the performance of the Grandpa Ranger module.


3h4x (changes requested):
Would be nice to have a test for that.

3h4x (approved):
🤖

PR #1424, opened 2023-11-02
chore: add prettier rules and reformat the whole project

gvko:
* add `prettier` rules that will match the rules we've got in the other NodeJS projects
* re-format the whole `cli` project based on the new rules


3h4x (approved):
🪨

PR #1429, opened 2023-11-02
fix: cli reset register light client

gvko:
* add up-to-date `prettier` config and re-format the whole project
* fix the `signAndSendSafe` function from the t3rn SDK

### Summary by CodeRabbit

- Refactor: Updated the project's formatting with a new `prettier` configuration and reformatted the entire codebase for consistency.
- Bug Fix: Enhanced error handling in the `signAndSendSafe` function across the SDK to handle dispatch errors and error events in custom modules. This change improves reliability and provides more detailed transaction status information.
- Refactor: Improved the `Tx` class functions to handle errors more effectively and return appropriate values, enhancing the robustness of the code.


client/packages/sdk/.prettierrc
```diff
@@ -1,3 +1,6 @@
 {
+  "tabWidth": 2,
+  "useTabs": false,
+  "singleQuote": true,
```
MaciejBaj, line 4:
```suggestion
  "singleQuote": false,
```

MaciejBaj: 😇

client/packages/sdk/src/circuit/tx.ts
```diff
@@ -42,7 +42,8 @@ export class Tx {
     if (this.exportMode) {
       exportObj = new ExtrinsicExport(tx, this.signer.address)
     }
-    return new Promise((resolve, reject) =>
+
+    const result = await new Promise((resolve, reject) =>
```
3h4x, line 46: 👍

3h4x (approved):
lgtm

client/packages/sdk/.prettierrc
gvko: We already have 2 or 3 other projects who use single-quote with the prettier config. And the most widely-used standard in JS is single-quote.
I don't rly care so much about which quotes we use, but rather that we stick to industry standards and also across projects.
I'll put double quotes for this one, if you prefer.

PR #1431, opened 2023-11-03
test: extend mock single vacuum order to test all registered gateways

MaciejBaj:
Test warm up asset transfers in Roc to all reqistered gateways via Vacuum::singleOrder. 

CLI can be started with `pnpm cli mockWriter --as-sequential-tx` and with 1min breaks sends test orders for asset = 1 (rococo) out of Alice account. 
Screenshot 2023-11-03 at 07 30 54


MaciejBaj:
Screenshot 2023-11-03 at 06 56 23

PR #1503, opened 2023-12-22
fix: update automated XCM transactions signer to use CIRCUIT_SIGNER_KEY

chexware:
### Summary by CodeRabbit

- Chore: Introduced a delay in the CLI command 'xcmTransfer'. After executing the transfer, the system will now pause for 10 seconds before displaying a success message and stopping the progress indicator. This change enhances user experience by providing a clear indication when the operation is fully completed.


3h4x (approved):
lgtm

PR #1566, opened 2024-02-19
feat: connect Vacuum single order as EVM Precompile

MaciejBaj:
### Summary by CodeRabbit

- New Feature: Introduced EVM Precompile functionality for Vacuum single orders, enhancing interoperability between EVM and Vacuum.
- New Feature: Added `VacuumAccess` trait with methods for handling EVM orders, improving the modularity of order processing.
- New Feature: Implemented a new struct `VacuumEVMOrder` for creating and decoding RLP-encoded Vacuum orders, providing a structured way to handle these orders.
- Refactor: Updated configuration in `pallet_3vm` module to include `VacuumEVMApi` and modified `AddressMapping` to use `EvmAddressMapping<Runtime>`, aligning with the new EVM-related features.
- New Feature: Integrated a new pallet named `AccountMapping` into the runtime, expanding the capabilities of account management.


chexware (commented):
Looks good, I introduced changes  to the runtime-level setup for precompiles so I will update the setup to be compatible with them before merging

chexware (approved):

runtime/t2rn-parachain/src/precompiles.rs
```diff
@@ -0,0 +1,57 @@
...
+        (
+            sp_core::H160([7u8; 20]),
+            evm_precompile_util::KnownPrecompile::ECRecoverPublicKey
+        ),
+        (
+            sp_core::H160([8u8; 20]),
+            evm_precompile_util::KnownPrecompile::Portal
+        ),
+        (
+            sp_core::H160([9u8; 20]),
```
chexware, line 44: This address overlaps with TokenPrecompile addresses and that will produce errors... We could just route 
it to the desired address in the new T3rnPrecompiles setup instead...

runtime/t0rn-parachain/src/precompiles.rs
```diff
@@ -0,0 +1,57 @@
...
+        (
+            sp_core::H160([7u8; 20]),
+            evm_precompile_util::KnownPrecompile::ECRecoverPublicKey
+        ),
+        (
+            sp_core::H160([8u8; 20]),
+            evm_precompile_util::KnownPrecompile::Portal
+        ),
+        (
+            sp_core::H160([9u8; 20]),
```
chexware, line 44: Same issue as T2rn... this address overlaps with the addresses linked to TokenPrecompile. We should use the T3rnPrecompile setup and route the address from there (with or without using hash function)

chexware (changes requested):
Left 2 comments... Basically we should keep T3rnPrecompiles setup as it allow us to link all AssetIds to TokenPrecompile... We can easily replace the hash function with H160 addresses if that is the preferred approach

runtime/t0rn-parachain/src/precompiles.rs
chexware: Fixed in [`271c1ed` (#1566)](https://github.com/t3rn/t3rn/pull/1566/commits/271c1edc2264e4eb56949cd7b0851f0d5530c3c9)

runtime/t2rn-parachain/src/precompiles.rs
chexware: Fixed in [`271c1ed` (#1566)](https://github.com/t3rn/t3rn/pull/1566/commits/271c1edc2264e4eb56949cd7b0851f0d5530c3c9)

chexware (approved):

